PR #84, opened 2026-06-13
fix: close diff fences during truncation

### LINKED ISSUE #21: fix(review): diff truncation cuts mid-hunk and drops the closing code fence
### Description

When a diff exceeds `thrillhousebot.review.max-diff-lines`, the formatter truncates the last file's section by raw line count, which can cut in the middle of a hunk and drop the closing ```` ``` ```` fence of the ```` ```diff ```` block. The prompt sent to the model then has an unterminated code fence, so trailing prompt content can be misread as diff.

`src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewDiffFormatter.java` — `truncateSection()` (around line 294) slices `section.split("\n")` and appends a "(patch truncated …)" line without re-closing the fence opened in `formatFileSection()`.

### Steps to Reproduce

1. Set a small `max-diff-lines` (or open a PR whose diff exceeds the default 5000).
2. Ensure truncation lands inside the last included file's ```` ```diff ```` block.
3. Inspect the generated prompt / model input.

### Expected Behavior

Truncation happens at a hunk boundary where possible, and always closes the ```` ```diff ```` fence so the prompt stays well-formed.

### Actual Behavior

The closing fence is dropped mid-hunk; the diff block is left open in the prompt.

### Environment

- App version: v0.1.0

### Code of Conduct

- [x] I have searched for existing issues and this is not a duplicate.

rupayon123:
Fixes #21.

This keeps truncated diff sections from leaving an open Markdown code fence when the line budget cuts through a patch. If the retained lines are inside a fence, the formatter now reserves space to close it before adding the truncation marker.

I also added regression coverage for truncating inside a ```diff block and for the tight-budget helper case.

Check run locally:
- `git diff --check`

I could not run `./mvnw -Dtest=ReviewDiffFormatterTest test` on this machine because there is no Java runtime installed.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewDiffFormatter.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewDiffFormatter.java
@@ -325,8 +325,13 @@ public class ReviewDiffFormatter {
   private static boolean isInsideCodeFence(String[] lines, int lineCount) {
     var insideFence = false;
     for (var i = 0; i < lineCount; i++) {
-      if (lines[i].trim().startsWith("```")) {
-        insideFence = !insideFence;
+      var line = lines[i].trim();
+      if (line.startsWith("```")) {
+        if (!insideFence) {
+          insideFence = true;
+        } else if (line.substring(3).trim().isEmpty()) {
+          insideFence = false;
+        }
       }
     }
     return insideFence;
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewDiffFormatter.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewDiffFormatter.java
@@ -303,15 +303,33 @@ public class ReviewDiffFormatter {
       return "(patch truncated)\n";
     }
 
-    var sb = new StringBuilder();
     var contentLines = maxLines - 1;
+    if (isInsideCodeFence(lines, contentLines)) {
+      contentLines = Math.max(0, maxLines - 2);
+    }
+
+    var closesFence = isInsideCodeFence(lines, contentLines);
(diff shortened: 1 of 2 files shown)


rupayon123:
Pushed an update for the fence parsing case. The helper now treats ```diff as an opener, but only a plain fence line as the closer once it is already inside a block.

I left the omitted-line assertion as-is after checking the trailing blank line behavior from `split("\n", -1)`. `git diff --check` passes; I still can’t run Maven locally here because there is no Java runtime installed.

devops-thiago:
Thanks for working on this, and for adding the tests plus the git diff --check run.

The issue is that #83 already fixed the same bug ([#21](https://github.com/devops-thiago/ThrillhouseBot/issues/21), where truncateSection() left an open ```diff fence). It merged earlier today as b82217a. Both PRs rewrite the same lines of ReviewDiffFormatter.truncateSection(), so this branch now conflicts with main.

The merged version went a bit further, which is why I don't think it can be layered on top:

It cuts at @@ hunk boundaries, so a partial trailing hunk gets dropped whole instead of shown half-formed.
It fixes the omitted-line count on the fence-less path by using lineCount(section) instead of lines.length, which over-counts the empty trailing element from split("\n", -1).
It falls back to a header plus notice on very small budgets rather than opening a fence it can't close, and it handles a section that is missing its closing fence.
It matches the closing fence with an exact equals("```"), so an unchanged bare ``` context line (which shows up in diffs of Markdown files) isn't mistaken for the end of the patch.
One thing your PR tests that the merged tests don't is the shouldNotTreatFencedPatchContentAsClosingFence case, a Markdown fence inside the diff body. Main handles it correctly already, but I'll add a regression test for it so we keep that coverage.

Going to close this as a duplicate of #83. Thanks again for the contribution.

PR #204, opened 2026-06-21
perf(review): pre-filter human-to-human inline replies before the paginated fetch + auth

devops-thiago:
## What type of PR is this?

- [x] 🚀 Performance
- [x] ✅ Test

## Description

On a watched PR, **every** inline review-comment reply (`comment.inReplyToId != null`) was dispatched to `MaintainerReplyService`, which then ran the authorizer (a collaborator-permission API call) **and** walked up to 10 paginated pages of the review-comments API — only *then* discovering the thread is human-to-human with no `@thrillhousebot` mention, and posting nothing. So two maintainers chatting on a human-started inline thread cost ~1 auth round-trip + up to 10 list calls per reply, for zero output — meaningful GitHub REST rate-limit and cost amplification on busy PRs.

This adds a cheap pre-filter to the review-thread path:

- New `GitHubReviewClient.getReviewComment(...)` — a single `GET /repos/{owner}/{repo}/pulls/comments/{id}`.
- For a reply that does **not** mention the bot, `MaintainerReplyService` now fetches just the thread **root** comment and decides `rootIsBot` from that one call, bailing **before** the authorizer round-trip and the paginated list. Mentions and confirmed bot-threads proceed exactly as before.
- Authorization moved into an `authorized()` helper, consulted only once we've decided we'd actually post.
- The pre-filter fails soft: a root that can't be fetched is treated as not-the-bot's, so the reply is left alone.

**Net for a human-to-human reply: ~11 GitHub calls + an auth round-trip collapse to a single cheap GET, and the authorizer is never consulted.** The remaining cost is the queued executor task itself — eliminating that would require a blocking GitHub call on the webhook ACK path (which must return within 10s), so it stays in the async handler.

Side benefit: an unmentioned reply on a bot thread whose root sits **beyond** the 10-page list bound is now still answered (the single GET resolves it directly) instead of being silently dropped.

## Related Issues

N/A

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

`MaintainerReplyServiceTest` (26/26), `MaintainerReplyDispatcherTest` (2/2), and `WebhookControllerTest` (58/58) pass locally. Run with JaCoCo disabled per the known SMB-mount file-locking issue on this checkout; CI runs full coverage on local disk:

```
./mvnw -o test -Djacoco.skip=true -Dquarkus.jacoco.enabled=false \
  -Dtest='MaintainerReplyServiceTest,MaintainerReplyDispatcherTest,WebhookControllerTest'
```

New/updated tests:
- `unmentionedReplyOnHumanThreadSkipsBeforeListingAndAuthorizing` — only the single GET runs; no list, no authorizer, nothing posted.
- `preFilterFailSoftSkipsWhenRootFetchThrows` — a failed root GET is treated as non-bot and skips.
- `unmentionedReplyWithUnresolvableRootPostsNothing`, `mentionWhereListRootHasNullAuthorCitesNoFinding`, plus split mention vs. review-thread unauthorized cases.

## Checklist

## Screenshots / Logs

N/A

## Additional Notes

The source comments deliberately omit an issue/PR number (none was fabricated). No behavior change for mentions or for replies on the bot's own finding threads — only human-to-human, no-mention replies are now short-circuited earlier and more cheaply.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/github/GitHubReviewClient.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/github/GitHubReviewClient.java
@@ -72,6 +72,21 @@ public interface GitHubReviewClient {
       @QueryParam("per_page") int perPage,
       @QueryParam("page") int page);
 
+  /**
+   * Fetches a single review comment by id. A cheap one-call alternative to paging the whole
+   * review-comment list when a caller only needs the one comment (e.g. resolving a thread's root
+   * author before deciding whether heavier work is warranted).
+   */
+  @GET
+  @Path("/repos/{owner}/{repo}/pulls/comments/{commentId}")
+  @Produces(MediaType.APPLICATION_JSON)
+  PullRequestComment getReviewComment(
+      @HeaderParam("Authorization") String auth,
+      @HeaderParam("Accept") String accept,
+      @PathParam("owner") String owner,
+      @PathParam("repo") String repo,
+      @PathParam("commentId") long commentId);
+
   @POST
   @Path("/repos/{owner}/{repo}/pulls/{pullNumber}/comments")
   @Produces(MediaType.APPLICATION_JSON)
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/MaintainerReplyService.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/MaintainerReplyService.java
@@ -121,22 +121,10 @@ public class MaintainerReplyService {
   @ActivateRequestContext
   public void handle(ReplyTask task) {
     try {
-      if (!authorizer.isAuthorized(
-          task.owner(),
-          task.repo(),
-          task.installationId(),
-          task.commenterLogin(),
-          task.authorAssociation())) {
-        Log.infof(
-            "Ignoring conversational reply request from @%s on %s/%s #%d — not authorized",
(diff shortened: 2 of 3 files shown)


devops-thiago:
Closing as superseded by #202.

After the rebase onto post-#202 `main`, the optimization here is unreachable. #202 gated the `pull_request_review_comment` path on an explicit `@`-mention only and now builds every `ReplyTask` with `mentioned = true` (`WebhookController.java:400`). So in `handleReviewThreadReply` the new pre-filter:

```java
if (!task.mentioned() && !rootCommentIsBot(auth, task)) { ... }
```

can never evaluate `rootCommentIsBot(...)` — `!task.mentioned()` is always `false`, the `&&` short-circuits, and the new `getReviewComment` GET is never issued in production. The human-to-human replies this PR optimizes are already dropped at the webhook by #202 (zero downstream calls, versus the one GET here), so there is no remaining cost to save.

The new tests pass only because they construct `mentioned = false` tasks directly — a shape the webhook can no longer emit — so they exercise code that is unreachable in production (green over dead code).

A rebase can't reconcile this because the conflict is intent, not text: #204 wants to *answer* unmentioned replies on the bot's own finding threads, while #202 deliberately decided *not* to. If we want that behavior back, it should be a single PR that flips the `WebhookController` gate **and** carries this cheap pre-filter (reverting part of #202), rather than landing the downstream half on its own.

Happy to reopen/rework if we decide to revisit #202's mention-only scope.

Issue #216, opened 2026-06-23
fix(review): require a contiguous run for a FULL multi-line quote match


devops-thiago:
### Description

`FindingQuoteValidator.matchQuote` (`src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingQuoteValidator.java` ~line 399) judges a quoted `suggestion_old` as fully present by **per-line set membership** (`quoted.stream().filter(diffLines::contains).count() == quoted.size()`), not by a contiguous in-order run. `checkAmbiguity` (~line 407) only scrutinizes quotes that *do* match contiguously and deliberately leaves a non-contiguous match as `FULL`. So a fabricated multi-line quote whose individual lines each exist **scattered** across the diff (never together/in order) passes anti-fabrication validation and the finding is kept with its suggestion.

Single-line quotes (the common case) are unaffected; the multi-line suggestion-apply path (`resolveSuggestionRange`) independently requires contiguity, so the practical impact is mainly that a likely-hallucinated finding is treated as well-grounded rather than demoted.

### Change required

- For a multi-line quote, require a **contiguous in-order run** for the `FULL` verdict (reuse the existing `contiguousMatchStarts`/`matchesAt` machinery), demoting a scattered match to `PARTIAL` (or the agreed downgrade).

### Acceptance criteria

- A unit test: a 2+ line quote whose lines exist only scattered (non-adjacent) is NOT `FULL`; a genuine contiguous quote still is.
- `./mvnw verify` green; coverage holds.

### Environment

- `src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingQuoteValidator.java` (`matchQuote` ~395-403, `checkAmbiguity` ~407)

### Related

- Would be surfaced by #123 (adversarial validator stress-tests).

### Code of Conduct

- [x] I have searched for existing issues and this is not a duplicate.

PR #227, opened 2026-06-23
test(review): make the quote-verdict switch fully covered (follow-up #216)

devops-thiago:
## What type of PR is this?

- [x] ✅ Test
- [x] 🔧 Refactor

## Description

The enum `switch` in `FindingQuoteValidator.validate()` had no `default`, so javac emits a **synthetic default** branch that no test can reach — JaCoCo reports the switch line as *partially* covered. #216 changed that line, so the partial now sits on `main` and would count against patch coverage on the next PR that modifies the area.

Fix: fold the keep verdicts (`FULL`, `NO_QUOTE`) into an explicit `default` placed last. It's reached by the existing FULL/NO_QUOTE tests, so there's no synthetic default and the line is fully covered. `default`-keep is also the safe handling for any future verdict (it surfaces the finding rather than silently dropping it). **No behavior change.**

## Related Issues

Follow-up to #216. N/A (no separate issue).

## How Has This Been Tested?

- [x] Unit tests

- JaCoCo now reports **no partial branches** in `FindingQuoteValidator`; the existing `FindingQuoteValidatorTest` already covers all four arms (FULL/NO_QUOTE via keep tests, PARTIAL, AMBIGUOUS, NONE).
- Full suite: **1221 passing**, spotbugs + spotless clean (JDK 25).

## Checklist

- [x] I have added tests that prove my fix is effective (coverage verified via JaCoCo)
- [ ] I have updated the documentation accordingly (no doc change needed)

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingQuoteValidator.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingQuoteValidator.java
@@ -69,20 +69,6 @@ public class FindingQuoteValidator {
     var changed = false;
     for (ReviewResponse.Finding finding : response.findings()) {
       switch (classifyQuote(finding, index)) {
-        case FULL, NO_QUOTE -> {
-          if (descriptionCitesAbsentCode(finding, index)) {
-            Log.infof(
-                "Finding '%s' (%s:%d) cites code in its description that appears nowhere in the diff —"
-                    + DEMOTION_SUFFIX,
-                finding.title(),
-                finding.file(),
-                finding.line());
-            kept.add(withoutSuggestion(finding));
-            changed = true;
-          } else {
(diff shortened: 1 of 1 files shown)


src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingQuoteValidator.java
```diff
@@ -69,20 +69,6 @@ public ReviewResponse validate(ReviewResponse response, String diff) {
     var changed = false;
     for (ReviewResponse.Finding finding : response.findings()) {
       switch (classifyQuote(finding, index)) {
-        case FULL, NO_QUOTE -> {
```
devops-thiago, line 72: @thrillhousebot, what's the impact of removing these case in switch?

Issue #528, opened 2026-08-11
Concise-cap refusal message names an inert remedy under a per-model buffer override


devops-thiago:
From deep-audit round 4 (AUDIT4-A finding 1; introduced by #520 — rework, message polish only).

The concise-cap refusal message advises "raise REVIEW_OUTPUT_BUFFER_TOKENS to cover it" even when the effective buffer comes from a per-model `output-buffer-tokens` override — where the global env var is never read, so following the remedy re-boots into the identical refusal. Proven with an executable probe (override 4096, concise 8192).

Fix: when the override is the source of the effective buffer, name `thrillhousebot.ai.models."<model>".output-buffer-tokens` in the remedy instead.

devops-thiago:
Already fixed on `release/v0.6.0` — closing as completed.

**Fixing PR:** #529
**Merge commit:** `4ceb24aa` — confirmed an ancestor of the `release/v0.6.0` tip `fda4bc7` (`git merge-base --is-ancestor 4ceb24aa fda4bc7`).

**Verified in the tree** (not taken from the PR description):
- `StartupConfigValidator` computes `bufferKnob` as `thrillhousebot.ai.models."<model>".output-buffer-tokens` when a per-model override exists and only otherwise falls back to `REVIEW_OUTPUT_BUFFER_TOKENS`, interpolating it into the refusal message — so the message no longer names an inert remedy.

---

**Why this issue stayed open.** GitHub only auto-closes a `Fixes #N` / `Closes #N` reference when the pull request merges into the repository's **default** branch (`main`). All v0.6.0 work merges into `release/v0.6.0`, so the reference was recorded but never triggered the auto-close. The fix is in the branch; only the bookkeeping was missing.

Closed as part of the pre-v0.6.0 housekeeping sweep.

PR #542, opened 2026-08-11
fix(review): name the cap that applies and soft-fail an absent verifier body

### LINKED ISSUE #533: Truncation warning names the active model's cap for concise-model calls
Found while testing the v0.6.0 image in dev (PR #532 review).

Production log:
```
WARN [FindingVerificationService] Finding verification — keeping unverified findings.
Finding verification stopped at the model's response-length cap (finish_reason=length) ...
Raise the active model's max-output-tokens, or leave it unset to use the provider default.
```

`FindingVerifier` is `@RegisterAiService(modelName = "concise")`, so the knob that actually applies is `REVIEW_CONCISE_MAX_OUTPUT_TOKENS` — raising the active model's `max-output-tokens` does nothing for it.

Cause: `AiResponses.textOrThrowOnTruncation` builds `AiResponseTruncatedException` with the single-arg constructor, so `conciseModelImplicated` is always false and every blocking assistant gets the active-model wording. The streaming summary lane passes the flag correctly; the blocking helper never does.

Same class as #528 (a remedy naming a knob that does not apply). Log-only for the verifier; the other blocking assistants share the helper, so check whether any surface it to users.

Fix direction: thread the concise flag through the helper from the calling assistant, with a test per lane.

### LINKED ISSUE #534: Verifier NPEs when the model returns a null/empty content body
Seen in dev during dogfood testing (ThrillhouseBot-test #7-#10 run):

```
WARN [FindingVerificationService] Finding verification failed — keeping unverified findings:
java.lang.NullPointerException: Cannot invoke "String.strip()" because ...
  at ReviewResponseParser.extractJson(ReviewResponseParser.java:228)
```

`FindingVerificationService` passes the unwrapped body straight into `extractJson`:

```java
var raw = AiResponses.textOrThrowOnTruncation(result, "Finding verification");  // may return null
var verdicts = mapper.readValue(ReviewResponseParser.extractJson(raw), VerificationResponse.class);
```

`extractJson` opens with `raw.strip()`, so a null body NPEs. The fail-open catch keeps the findings, so impact is limited to noise plus a lost verification pass — but it turns an expected "no response" into an exception path.

Attribution: the unguarded call is pre-existing (present in v0.5.0). What is new is `AiResponses.textOrThrowOnTruncation` (#504), whose javadoc states "A null result passes through as null: the callers already treat no response as a soft failure" — that contract is false for this caller.

Likely trigger: with `REVIEW_CONCISE_MAX_OUTPUT_TOKENS` unset and reasoning at max effort, the whole output budget can be consumed by reasoning tokens, leaving zero content tokens — so the body is empty/null without necessarily reporting finish_reason=length.

Fix direction: null-guard at the caller (treat as the documented soft failure and keep unverified findings), or make `extractJson` null-safe. Also correct the helper javadoc or make every caller honor it.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix

## Description

Two defects in the blocking response-unwrap lane, in the same files.

**#533 — the truncation remedy named a knob that cannot affect the call.** `AiResponses.textOrThrowOnTruncation` always built `AiResponseTruncatedException` with the single-arg constructor, so `conciseModelImplicated` was always false and every blocking lane got the active-model wording. Two of the three callers are bound to the `concise` named model, whose cap is `REVIEW_CONCISE_MAX_OUTPUT_TOKENS` and which deliberately never receives the active model's `max-output-tokens` (`ChatModelCustomizers` applies every shared parameter to it *except* the response cap). Operators of those lanes were told to raise a setting with no effect on the call that failed — the same class of defect as #528.

Every caller now states its lane through a new `AiResponses.ModelLane`, which carries the remedy text and sets the concise flag. There is no defaulting overload: a `Result` carries no trace of the model that produced it, and the implicit default is exactly what made this bug silent.

Caller classification (all three checked):

| Caller | Assistant | Binding | Lane |
| --- | --- | --- | --- |
| `AbstractPrSuggestionGenerator#callAssistant` (`/describe`, `/improve`, `/changelog`, `/generate-tests`, `/add-docs`) | `PrDescribeAssistant`, `PrImproveAssistant`, `ChangelogAssistant`, `UnitTestAssistant`, `DocGenerator` | `@RegisterAiService` (default) | `ACTIVE` |
| `MaintainerReplyService#generateReply` | `ReplyAssistant` | `@RegisterAiService(modelName = "concise")` | `CONCISE` |
| `FindingVerificationService#verify` | `FindingVerifier` | `@RegisterAiService(modelName = "concise")` | `CONCISE` |

No blocking lane surfaces the message to users — all three log it and post nothing / keep findings — so this is operator-log copy only. The streaming summary lane already marked itself via `implicatingConciseModel()` and is unchanged.

**#534 — the verifier NPE'd on an absent body.** `FindingVerificationService` passed the unwrapped body straight into `ReviewResponseParser.extractJson`, which opens with `raw.strip()`. With the whole output budget spent on reasoning tokens the provider completes with no content and no length stop, so the body is null — and the helper's javadoc claim that "the callers already treat no response as a soft failure" was false for exactly this one. Checked every caller: the `/describe`-family generators (`suggestion == null || suggestion.isBlank()`) and the reply lane (`reply == null || reply.isBlank()`) did honour it; the verifier alone did not, and it is also the only lane that calls `mapper.readValue(extractJson(...))` inline instead of through a parser class that guards.

The verifier now keeps the unverified findings on an absent or blank body — its documented error contract — instead of routing an expected outcome through the fail-open catch as a verification fault. The helper's javadoc now states the contract and what each caller does with it, and `extractJson` yields `""` for an absent body rather than raising from inside the strip.

## Related Issues

Fixes #533
Fixes #534

## How Has This Been Tested?

- [x] Unit tests

### Red/green — #534, the NPE (red on the unmodified base `abd76e5`)

`ReviewResponseParserTest#shouldYieldEmptyExtractionForAnAbsentBody` reproduces the production frame verbatim:

```
[ERROR] dev.thiagogonzaga.thrillhousebot.review.ai.ReviewResponseParserTest.shouldYieldEmptyExtractionForAnAbsentBody -- Time elapsed: 0.195 s <<< ERROR!
java.lang.NullPointerException: Cannot invoke "String.strip()" because "raw" is null
	at dev.thiagogonzaga.thrillhousebot.review.ai.ReviewResponseParser.extractJson(ReviewResponseParser.java:228)
	at dev.thiagogonzaga.thrillhousebot.review.ai.ReviewResponseParserTest.shouldYieldEmptyExtractionForAnAbsentBody(ReviewResponseParserTest.java:224)
```

`FindingVerificationServiceTest#keepsUnverifiedFindingsWithoutParsingWhenTheModelReturnsNoBody` pins the caller. The findings survive either way (the fail-open catch caught the NPE), so it asserts the absent body never reaches the extraction:

```
[ERROR] dev.thiagogonzaga.thrillhousebot.review.ai.FindingVerificationServiceTest.keepsUnverifiedFindingsWithoutParsingWhenTheModelReturnsNoBody -- Time elapsed: 1.232 s <<< FAILURE!
org.mockito.exceptions.verification.NeverWantedButInvoked:

ReviewResponseParser.class.extractJson(
    <any>
);
Never wanted here:
-> at dev.thiagogonzaga.thrillhousebot.review.ai.ReviewResponseParser.extractJson(ReviewResponseParser.java:228)
But invoked here:
-> at dev.thiagogonzaga.thrillhousebot.review.ai.FindingVerificationService.verify(FindingVerificationService.java:118) with arguments: [null]
```

### Red/green — #533, the knob named per lane

The pre-fix API cannot express a lane, so the red run keeps the new signature and restores the pre-fix body (the single-arg constructor with the active-model wording) — i.e. exactly the code this PR replaces. `AiResponsesTest#namesTheConciseCapForACallOnTheConciseModel`:

```
[ERROR] dev.thiagogonzaga.thrillhousebot.review.ai.AiResponsesTest.namesTheConciseCapForACallOnTheConciseModel -- Time elapsed: 0.055 s <<< FAILURE!
org.opentest4j.AssertionFailedError: the message must name the knob that actually caps this lane: Finding verification stopped at the model's response-length cap (finish_reason=length), so the response is incomplete. Raise the active model's max-output-tokens, or leave it unset to use the provider default. ==> expected: <true> but was: <false>
```

`ConciseModelTruncationTest` drives the real concise-bound AI services against mocked `@ModelName("concise")` model beans, and both lanes reproduce the reported log line:

```
[ERROR]   ConciseModelTruncationTest.aLengthStopOnTheConciseVerifierCallRaisesTheTruncationError:100 the verifier runs on the concise model, so its cap is the one to name: Finding verification stopped at the model's response-length cap (finish_reason=length), so the response is incomplete. Raise the active model's max-output-tokens, or leave it unset to use the provider default. ==> expected: <true> but was: <false>
[ERROR]   ConciseModelTruncationTest.aLengthStopOnTheConciseReplyCallRaisesTheTruncationError:120 the reply lane runs on the concise model, so its cap is the one to name: Maintainer reply stopped at the model's response-length cap (finish_reason=length), so the response is incomplete. Raise the active model's max-output-tokens, or leave it unset to use the provider default. ==> expected: <true> but was: <false>
```

The `ACTIVE` lane is pinned in the other direction: `namesTheActiveModelsCapForACallOnTheDefaultModel` asserts the default-model message keeps the active-model remedy, never mentions `REVIEW_CONCISE_MAX_OUTPUT_TOKENS`, and leaves the concise flag false.

### Gates

- `./mvnw -B spotless:apply` then `./mvnw -B clean compile spotbugs:check spotless:check` — BUILD SUCCESS, `BugInstance size is 0`
- `./mvnw -B clean test` — Tests run: 2600, Failures: 0, Errors: 0, Skipped: 0
- Coverage: jacoco ∩ `git diff -U0 abd76e5...HEAD` over changed main code — 18 executable changed lines, 0 uncovered lines, 0 uncovered branches

## Checklist

## Additional Notes

Two call sites outside the nominal file scope were required and are one line plus a comment each: `MaintainerReplyService` and `AbstractPrSuggestionGenerator#callAssistant` are the other two callers of the helper, and #533 is not fixed for the reply lane without them. No behaviour outside the truncation message changed there.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/AbstractPrSuggestionGenerator.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/AbstractPrSuggestionGenerator.java
@@ -296,11 +296,17 @@ public abstract class AbstractPrSuggestionGenerator {
    * RuntimeException}, degrades to {@code null} (post nothing); a usable reply is returned
    * stripped. {@code command} labels the operation in logs (e.g. {@code "/describe"}). Subclasses
    * supply the actual call (and apply any command-specific post-filtering to the result).
+   *
+   * <p>Every subclass's assistant is bound to the default model ({@code /describe}, {@code
+   * /improve}, {@code /changelog}, {@code /generate-tests}, {@code /add-docs}), so a truncation
+   * here is capped by the active model's {@code max-output-tokens} — the concise-bound calls
+   * (summary, verifier, replies) do not come through this path.
    */
   protected String callAssistant(String command, Supplier<Result<String>> assistantCall) {
     try {
       String suggestion =
-          AiResponses.textOrThrowOnTruncation(assistantCall.get(), command + " assistant");
+          AiResponses.textOrThrowOnTruncation(
+              assistantCall.get(), command + " assistant", AiResponses.ModelLane.ACTIVE);
       if (suggestion == null || suggestion.isBlank()) {
         Log.debugf("%s assistant produced an empty suggestion — posting nothing", command);
         return null;
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/MaintainerReplyService.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/MaintainerReplyService.java
@@ -229,7 +229,10 @@ public class MaintainerReplyService {
                   // codeContext is the diff/hunk under discussion: fence it byte-exact.
                   PromptTemplateEscaper.fence(codeContext),
                   PromptTemplateEscaper.escape(thread)),
-              "Maintainer reply");
+              "Maintainer reply",
+              // ReplyAssistant is bound to the concise named model, so the cap that can cut this
+              // call is REVIEW_CONCISE_MAX_OUTPUT_TOKENS, not the active model's.
+              AiResponses.ModelLane.CONCISE);
       if (reply == null || reply.isBlank()) {
         Log.debug("Reply assistant produced an empty reply — posting nothing");
         return null;
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/AiResponses.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/AiResponses.java
@@ -33,14 +33,59 @@ public final class AiResponses {
   private AiResponses() {}
 
   /**
+   * Which model binding the unwrapped call runs on. The two bindings are capped by different knobs,
+   * and a truncation remedy only helps an operator when it names the one that actually applies — so
+   * every call site states its lane instead of inheriting a default. A {@link Result} carries no
+   * trace of the model that produced it, and the implicit "active model" wording this helper used
+   * to apply to every blocking call told operators of the concise-bound lanes to raise a cap that
+   * does not bound them.
+   */
+  public enum ModelLane {
+
+    /** The default binding, capped by the active model's {@code max-output-tokens}. */
+    ACTIVE(
+        "Raise the active model's max-output-tokens, or leave it unset to use the provider"
+            + " default."),
+
+    /**
+     * The {@code concise} named binding — the summary, verifier and reply calls — capped by {@code
+     * REVIEW_CONCISE_MAX_OUTPUT_TOKENS}. The active model's {@code max-output-tokens} is
+     * deliberately never applied to it (see {@link ChatModelCustomizers}), so naming that knob here
+     * would send the operator to a setting with no effect on this call.
+     */
+    CONCISE(
+        "This call runs on the concise named model, so raise REVIEW_CONCISE_MAX_OUTPUT_TOKENS (the"
+            + " active model's max-output-tokens does not cap it), or leave it unset to use the"
+            + " provider default.");
+
+    private final String remedy;
+
+    ModelLane(String remedy) {
+      this.remedy = remedy;
+    }
+
+    /** The operator instruction naming the response cap that applies to this lane. */
+    String remedy() {
+      return remedy;
+    }
+  }
+
+  /**
    * Returns the response text, or throws {@link AiResponseTruncatedException} when the model
    * stopped at its response-length cap. {@code what} names the call for the operator ("/improve",
-   * "Finding verification"), since the exception is what the caller logs.
+   * "Finding verification"), since the exception is what the caller logs; {@code lane} names the
+   * model binding the call ran on, so the message and the exception's {@link
+   * AiResponseTruncatedException#conciseModelImplicated() concise flag} point at the cap that
+   * actually cut it.
    *
-   * <p>A {@code null} result passes through as {@code null}: the callers already treat "no
-   * response" as a soft failure, and that is not the same condition as a truncation.
+   * <p>A {@code null} result, and a completed response with no content body, both pass through as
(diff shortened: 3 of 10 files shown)


devops-thiago:
Addressing the review's low-confidence note ("ACTIVE lane classification relies on @RegisterAiService bindings not shown in the diff") — fair to raise, since the bindings genuinely are not in this diff. Verified against every AI service in the repo:

| Assistant | `modelName` | Lane assigned here |
|---|---|---|
| `PrDescribeAssistant` | none (active) | ACTIVE |
| `PrImproveAssistant` | none (active) | ACTIVE |
| `ChangelogAssistant` | none (active) | ACTIVE |
| `UnitTestAssistant` | none (active) | ACTIVE |
| `DocGenerator` | none (active) | ACTIVE |
| `ReplyAssistant` | `concise` | CONCISE |
| `FindingVerifier` | `concise` | CONCISE |
| `PrSummarizer` | `concise` | streaming lane, self-marks, untouched |
| `PrReviewer` | none (active) | not a caller of this helper |

All five subclasses reached through `callAssistant` are active-model services, and the two `CONCISE` classifications are the only concise-bound callers of the helper. No change needed.

Context on impact: the remedy text is operator-log copy only — no blocking lane surfaces it to users — so a misclassification would have produced a confusing log line rather than a broken run.

PR #544, opened 2026-08-11
fix(review): fill the walkthrough file summaries on large PRs

### LINKED ISSUE #536: Walkthrough file summaries collapse to "-" on large PRs
Reported from production use and reproduced by comparing PR sizes.

**Large PR (devops-thiago/ThrillhouseBot#532):** 100 changed files, 20 walkthrough rows rendered, and **all 20 rows show `-` in the Summary column** — `file_summaries` came back entirely absent, so the file-by-file walkthrough carries no information at exactly the PR size where it matters most.

**Small PRs (ThrillhouseBot-test #7-#10, 10-17 files):** every row carries a real one-line summary. Only the renamed row shows `-`.

Two things to settle:

1. **Root cause of the collapse.** Either the model omits `file_summaries` on large/multi-call reviews, or it is returned mis-shaped and dropped in parsing (the #510 tolerance work would swallow a mis-shaped field silently). Decode the raw response for a large PR before choosing a fix. Note the prompt caps the array at 15 entries while the walkthrough renders more rows than that, so even a correct response cannot fill every row — the cap and the render width should agree.

2. **Rename rows should not render a bare dash.** A pure rename is deliberately excluded from AI review, so no summary exists for it by design. Rendering `-` makes correct behavior look like a failure (it is what prompted this report). Render something explicit such as "Pure rename — content unchanged".

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix

## Description

On a large PR the file-by-file walkthrough rendered every row with `-`. Diagnosis first, then three fixes.

### Root cause

The dashboard/DB was not reachable from this machine, so the diagnosis is read off the artifact the run actually posted plus the parser. The summary comment on #532 (169 files) shows:

- `pr_purpose` rendered → the `summary` object mapped cleanly, so the `parseWithoutSummary` salvage (which drops the whole summary node) did not fire;
- the Control-Flow Diagram rendered → `walkthrough_diagram`, which the summary prompt lists **after** `file_summaries`, was present, so the response was not cut short before reaching the field.

The only remaining explanation is that the multi-call summary pass **omitted `file_summaries` outright** — and its prompt gives it every reason to. `SUMMARY_SYSTEM` tells that call "You do not see the diff", forbids inventing anything, and then asks for "a single line on what changed in that file", which the diff is the only source for. A model obeying both instructions emits nothing. That also explains the size split exactly: small PRs take the single-call path, whose `SYSTEM` prompt does see the diff, and their rows carry real summaries.

Two silent drops sat behind it in the parser, and both are live in production even though the shape is recoverable:

- an entry keyed `file`/`description` instead of `path`/`summary` maps — Quarkus's ObjectMapper ignores unknown properties — to a `FileSummary` with a **null path**, which `summariesByPath` filters out without a word, so every row shows `-`;
- a map or scalar where the array belongs fails schema mapping, at which point the #510 salvage discards the **whole** summary, taking `pr_purpose` and the description gaps with it.

### Fixes

1. **`SUMMARY_SYSTEM`**: `file_summaries` is now marked REQUIRED and grounded in the material that call actually holds — the authoritative changed-file list with statuses, ±counts and directory breakdown, the computed findings, and the PR description — with an explicit statement that a line at that granularity is wanted and is *not* invention, that inventing means naming behavior the material does not show, and that omitting the field blanks every row of the rendered table.
2. **Parser tolerance**: `normalizeFileSummaries` recovers the map form, key aliases (`file`/`filename`/`file_path`, `description`/`note`/`change`/`text`), `"path: summary"` strings, and a single entry emitted where the array belongs; entries carrying no usable pair are dropped and the count is logged instead of vanishing. Both prompts now also pin the key spelling and the array shape.
3. **Cap vs. render**: the prompt capped the array at 15 while the walkthrough renders 20 rows, so five rows could never be filled by any correct response. `PrSummaryGenerator.MAX_FILE_ROWS` now reads `PrReviewPrompts.MAX_FILE_SUMMARIES`, and a content test pins the prompt prose to the same constant.
4. **Pure renames**: a pure rename is excluded from AI review by design, so it never carries a model summary; its `-` read as a failure. The pure-rename flag now travels with the walkthrough row (`ChangedFile`) and the cell renders `Pure rename — content unchanged`. A rename that also changed content was reviewed, so it keeps the bare dash.

### Note on scope

`VerdictBuilder.toChangedFiles` is outside the assigned lane; it is the one place that builds `ChangedFile`, so the pure-rename flag cannot reach the renderer without it. The touch is confined to that three-line method (`ReviewDiffFormatter.isPureRename(f)` passed as the new component) and its Javadoc.

## Related Issues

Fixes #536

## How Has This Been Tested?

- [x] Unit tests

Red/green proof, parser tolerance (`ReviewResponseParser.java` reverted, tests kept):

```
[ERROR] Tests run: 45, Failures: 2, Errors: 4, Skipped: 0
[ERROR]   ReviewResponseParserTest.shouldRecoverFileSummariesKeyedFileAndDescription:489 expected: <src/A.java> but was: <null>
[ERROR]   ReviewResponseParserTest.shouldKeepTheSummaryWhenFileSummariesIsAScalar:577 the summary object must survive a mis-shaped field ==> expected: not <null>
[ERROR]   ReviewResponseParserTest.shouldRecoverFileSummariesEmittedAsAMapKeyedByPath:506 NullPointer Cannot invoke "dev.thiagogonzaga.thrillhousebot.review.ai.ReviewResponse$Summary.fileSummaries()" because the return value of "dev.thiagogonzaga.thrillhousebot.review.ai.ReviewResponse.summary()" is null
[ERROR]   ReviewResponseParserTest.shouldRecoverFileSummariesEmittedAsPathColonSummaryStrings:526 NullPointer Cannot invoke "...ReviewResponse$Summary.fileSummaries()" because the return value of "...ReviewResponse.summary()" is null
[ERROR]   ReviewResponseParserTest.shouldWrapASingleFileSummaryObjectEmittedWhereTheArrayBelongs:542 NullPointer Cannot invoke "...ReviewResponse$Summary.fileSummaries()" because the return value of "...ReviewResponse.summary()" is null
[ERROR]   ReviewResponseParserTest.shouldDropHalfEntriesButKeepTheUsableOnesAndTheSummary:560 NullPointer Cannot invoke "...ReviewResponse$Summary.fileSummaries()" because the return value of "...ReviewResponse.summary()" is null
```

The first line is the exact silent drop this issue reports: with the production (unknown-property-tolerant) mapper the entry survives Jackson with `path == null` and the renderer skips it. The rest are the shapes that cost the entire summary object.

Red/green proof, pure-rename cell (`summaryCell` reverted to `return "-";`):

```
[ERROR] PrSummaryGeneratorTest.shouldExplainPureRenameRowInsteadOfRenderingBareDash:637 ## 🤖 ThrillhouseBot PR Summary
...
### Changed Files
| File | Change | Summary |
|------|--------|---------|
| `src/Moved.java` | Renamed | - |
| `src/Edited.java` | Renamed | - |
```

Red/green proof, cap agreement (`MAX_FILE_SUMMARIES` forced back to 15):

```
[ERROR] PrReviewPromptsContentTest.bothPromptsAskForAsManyFileSummariesAsTheWalkthroughRenders:673->assertContains:40 the review prompt's file_summaries cap must be the walkthrough row bound (#536) — missing marker: "cap the array at 15 entries" ==> expected: <true> but was: <false>
[ERROR] PrSummaryGeneratorTest.changesOverviewUsesAuthoritativeTotalsWhenReviewableCountsDiverge:714 expected: <true> but was: <false>
```

Gates:

- `./mvnw -B clean compile spotbugs:check spotless:check` → `BugInstance size is 0`, BUILD SUCCESS
- `./mvnw -B clean test` → `Tests run: 2607, Failures: 0, Errors: 0, Skipped: 0`
- jacoco ∩ `git diff -U0 abd76e5...HEAD` over changed main code → 69 changed executable lines, zero uncovered lines and zero uncovered branches

## Checklist

## Additional Notes

The remaining prompt-side risk is that the model still under-fills the array on very wide PRs; the rollup row (`…and N more file(s)`) already covers what the table cannot show.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/PrSummaryGenerator.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/PrSummaryGenerator.java
@@ -16,6 +16,7 @@
 package dev.thiagogonzaga.thrillhousebot.review;
 
 import dev.thiagogonzaga.thrillhousebot.config.ThrillhouseConfig;
+import dev.thiagogonzaga.thrillhousebot.review.ai.PrReviewPrompts;
 import dev.thiagogonzaga.thrillhousebot.review.ai.ReviewResponse;
 import jakarta.enterprise.context.ApplicationScoped;
 import jakarta.inject.Inject;
@@ -64,11 +65,33 @@ public class PrSummaryGenerator {
   /**
    * Upper bound on rows in the changed-files walkthrough. Keeps the comment within GitHub's size
    * budget on large PRs; any files beyond this are rolled up into a trailing "… and N more" note.
+   *
+   * <p>Read from {@link PrReviewPrompts#MAX_FILE_SUMMARIES} so the number of rows rendered and the
+   * number of per-file summaries the prompt asks the model for cannot drift apart: a smaller prompt
+   * cap would guarantee rows no response could ever fill (#536).
    */
-  static final int MAX_FILE_ROWS = 20;
+  static final int MAX_FILE_ROWS = PrReviewPrompts.MAX_FILE_SUMMARIES;
 
-  /** One changed file in the walkthrough: its path and the diff's authoritative change type. */
-  public record ChangedFile(String path, String changeType) {}
+  /**
+   * Summary cell for a pure-rename row. A pure rename is excluded from AI review by design (its
+   * content is unchanged, so there is nothing to review), which means the model never returns a
+   * summary for it — rendering the bare "-" made that correct behavior read as a failed summary
+   * (#536), so the row states the reason instead.
+   */
+  static final String PURE_RENAME_SUMMARY = "Pure rename — content unchanged";
+
+  /**
+   * One changed file in the walkthrough: its path, the diff's authoritative change type, and
+   * whether it is a pure rename — renamed with no content change, hence never sent to the model and
+   * never carrying a model summary.
+   */
+  public record ChangedFile(String path, String changeType, boolean pureRename) {
+
+    /** A row whose pure-rename status is not distinguished; treated as ordinary content change. */
+    public ChangedFile(String path, String changeType) {
+      this(path, changeType, false);
+    }
+  }
 
   /**
    * Upper bound on the rendered Mermaid source. A diagram larger than this is dropped rather than
@@ -400,7 +423,7 @@ public class PrSummaryGenerator {
           .append("` | ")
           .append(changeTypeLabel(file.changeType()))
           .append(" | ")
-          .append(summary.isBlank() ? "-" : MarkdownSafe.tableCell(summary.strip()))
+          .append(summaryCell(file, summary))
           .append(" |\n");
     }
     int rowsShown = Math.min(changedFiles.size(), MAX_FILE_ROWS);
@@ -409,6 +432,17 @@ public class PrSummaryGenerator {
       sb.append("\n_…and ").append(overflow).append(" more file(s)._\n");
     }
     sb.append("\n");
+  }
+
+  /**
+   * The walkthrough's Summary cell. The model's note wins whenever it supplied one; otherwise a
+   * pure rename says why it has none, and every other unsummarized row keeps the bare dash.
+   */
+  private static String summaryCell(ChangedFile file, String summary) {
+    if (!summary.isBlank()) {
+      return MarkdownSafe.tableCell(summary.strip());
+    }
+    return file.pureRename() ? PURE_RENAME_SUMMARY : "-";
   }
 
   /** Indexes the model's per-file notes by path; a duplicate path keeps the first note. */
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/VerdictBuilder.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/VerdictBuilder.java
@@ -427,12 +427,17 @@ public class VerdictBuilder {
   }
 
   /**
-   * Projects the reviewed diff onto the (path, change type) rows the summary walkthrough renders.
+   * Projects the reviewed diff onto the (path, change type, pure-rename) rows the summary
+   * walkthrough renders. The pure-rename flag travels with the row so the walkthrough can say why
+   * such a file carries no model summary instead of rendering a bare dash (#536).
    */
   static List<PrSummaryGenerator.ChangedFile> toChangedFiles(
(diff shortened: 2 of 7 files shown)


devops-thiago:
Addressing the two findings the latest review still holds open. Neither has an inline thread (both were low-confidence "double-check" items), so answering here.

**1. LOW — "Prompt says mis-keyed file_summaries keys are silently discarded, but this PR's parser now recovers them" — accurate, fixed.**

Correct, and it is the same defect class this stack widens in dimension 4: a sentence contradicting the code beside it. Fixed in `c42c25a`, in both copies of the instruction (single-call and multi-call). The requirement stays — exact `path`/`summary` keys, array shape — only the false consequence is gone.

**2. MEDIUM — "Verify the summary prompt actually receives the changed-file list the new file_summaries bullet promises" — fair question, verified, no change needed.**

The multi-call summary call is assembled in `FindingPipeline.runMultiCall`:

```java
var overview = clampOverview(changedFilesOverview(ctx, plan), promptInputs);
var summaryInputs = new AiReviewService.SummaryInputs(
    promptInputs.prContext(),                                 // PR title and description
    escape(budgetedFindingsJson(refined.findings(), ...)),    // the computed findings
    escape(overview),                                         // the changed-file list
    promptInputs.previousFindings(), promptInputs.repoInstructions());
```

`changedFilesOverview` emits one row per file as `path (status, +added -deleted)` with coverage markers, preceded by the pure-rename rollup and a scope header carrying the totals and the per-directory breakdown. So every element the bullet names is present at runtime.

The finding's real point — that in-diff content tests only pin prose inside the constant and cannot prove the runtime prompt contains the list — is right about this diff, but that behaviour is already pinned elsewhere in the suite: `FindingPipelineTest` captures the actual `SummaryInputs` from the multi-call path and asserts on its content, including `"a.java (modified, +3 -0 — partially analyzed"` and `"(modified, +7 -0)\n"`.

**One residual limit worth stating**, since it bounds how completely this fixes the reported collapse: `clampOverview` truncates the overview to fit the prompt budget, and the rollup and scope header are deliberately placed first so clamping drops the tail. On a very large PR the per-file rows beyond the clamp are genuinely absent from the prompt, and the model cannot summarize what it was not given. The grounding sentence is true for every file inside the clamp; files past it remain unsummarized by construction, and closing that gap is a separate change.

Issue #569, opened 2026-08-11
A pure rename withheld from the model is reported as missing work in description_gaps


devops-thiago:
Found in round-2 dogfood scoring, on ThrillhouseBot-test #15 (java) and #13 (kotlin) — the same false claim in both.

Each PR renames `README.md` to `docs/<LANG>.md` as a pure rename and says so in its description. The review then produces, in one comment:

1. the banner — *"1 pure rename omitted from AI review (README.md → docs/JAVA.md)"*
2. a walkthrough row — `docs/JAVA.md | Renamed | Pure rename — content unchanged`
3. and, in **Description vs. Implementation**, a claim that the rename described in the PR body is **not present in the diff**

All three in the same posted comment, within a few lines of each other. The rename is correctly excluded from the model's input (that part is working as designed, #386), the model is not told it was excluded, and so it reports the withheld change as unimplemented.

This is the same root cause as #566 — absence from the provided material treated as absence from the PR — on a different surface. It is worse here in one respect: `description_gaps` is presented to the maintainer as a mismatch between what they claimed and what they delivered, so the reader is told to go and do work the PR already contains. It is the single most misleading output either review produced.

Fix direction is shared with #566: tell the model what was withheld and why (pure renames, ignore-listed paths), so it can distinguish "not in my material" from "not in the PR". Narrower stopgap: exclude claims about renamed-but-withheld files from `description_gaps` specifically, since the renderer already knows exactly which paths those are.

Credit: the corpus makes this reproducible — every PR in the round renames the README as a pure rename, so any of them can serve as a regression case.

devops-thiago:
Round-2 command scoring shows this is broader than the original report, and has a second-order harm I had not seen.

**Reach.** The auto-posted review summary carries the false claim on **8 of 8** corpus PRs, self-contradicting inside a single comment whose own header names the rename and whose file table lists it as `Renamed`. `/describe` repeats it independently on **4 of 8** (angular, zig, react, c).

**Second-order harm.** The suggested description that `/describe` offers **deletes the author's true rename sentence** on all 8 PRs — so a maintainer who accepts the suggestion removes an accurate statement about their own PR because the bot could not see the file it had itself withheld. Relatedly, `/changelog` never emits a `Changed` entry for the rename on any of the 8.

[...]

Issue #17, opened 2026-06-13
fix(dashboard): fail closed when GitHub App owner cannot be resolved


devops-thiago:
### Description

Dashboard access control fails **open**. When the GitHub App owner cannot be resolved and `thrillhousebot.dashboard.github.account-owner` is unset, `DashboardAccessChecker.hasAccess()` returns `true` for every authenticated GitHub user, exposing cost, token, and session data.

`src/main/java/dev/thiagogonzaga/thrillhousebot/dashboard/DashboardAccessChecker.java:96-99`: when `accountOwner()` is empty, `hasAccess()` returns `true`.

### Steps to Reproduce

1. Deploy without `thrillhousebot.dashboard.github.account-owner` set.
2. Put the app in a state where owner auto-resolution fails (the `GET /app` call errors or returns no owner login).
3. Log in to the dashboard with any GitHub account.
4. Observe full dashboard access.

### Expected Behavior

Fail closed: deny dashboard access and surface a clear "owner not configured / could not be resolved" message until `account-owner` is set.

### Actual Behavior

Access control is silently disabled; any successful GitHub login reaches the dashboard.

### Environment

- App version: v0.1.0

### Code of Conduct

- [x] I have searched for existing issues and this is not a duplicate.

Issue #15, opened 2026-06-13
test(frontend): add Vitest coverage for overview dashboard page


devops-thiago:
### Problem Statement

The dashboard frontend has Vitest coverage for `costs` and `tokens` pages, but **`overview/page.tsx` has no tests**. That page combines REST summary data with live WebSocket events — the most interactive screen — and regressions there are easy to miss.

### Proposed Solution

Add `frontend/app/(dashboard)/overview/page.test.tsx` following the pattern in `costs/page.test.tsx`:

1. Mock `@/lib/api` and return `mockSummary` from `@/lib/mock-data`.
2. Mock `@/hooks/useWebSocket` with controllable `events` and `connected` state.
3. Cover at least:
   - Renders summary stats after the API resolves (heading **Dashboard Overview**, key counts/cost from `mockSummary`).
   - Shows loading state while fetching.
   - Shows error + **Retry** when `api().summary()` fails, then recovers on retry.
   - Optional: one test with a `review.started` / `review.completed` event to assert live-review UI updates.

Run with `cd frontend && npm test`.

### Alternatives Considered

- E2E tests with Playwright — valuable later, but heavier than unit tests for a first contribution.
- Testing `buildLiveReviews` in isolation — would require exporting helpers; page-level tests match existing dashboard coverage.

### Priority

Important

### Additional Context

- Reference tests: `frontend/app/(dashboard)/costs/page.test.tsx`, `tokens/page.test.tsx`
- Mock fixtures: `mockSummary`, `mockSessionEvents` in `frontend/lib/mock-data.ts`
- Vitest setup: `frontend/vitest.config.ts`, `frontend/vitest.setup.ts`

- [x] I have searched for existing issues and this is not a duplicate.

Issue #11, opened 2026-06-13
release: end-to-end automated release chain — version-driven trigger, changelog, and the bump PR that started this issue


devops-thiago:
> _Scope widened 2026-08-13: originally this tracked one broken job (the post-release bump PR). That failure is a symptom of the release chain having manual seams; this issue now tracks closing all of them. The original bug report is preserved below as Part 1 — it remains the first deliverable, and PR #208 (fail loudly + RELEASING.md) is still the right immediate step._

### Goal

A release should require exactly **one human action** — declaring the new version — with everything else automated and auditable: tag creation, verification, publication, changelog, and the post-release version bump. Today the chain is `manual tag push → verify → scan → promote → release → bump-version`, with two manual seams: the operator creates the tag by hand at the start, and the bump PR fails at the end (Part 1).

### Part 1 — the original bug: bump-version PR creation fails (first deliverable)

The `bump-version` job in `.github/workflows/release.yml` pushes `chore/bump-*` but `gh pr create` fails with:

`GraphQL: GitHub Actions is not permitted to create or approve pull requests (createPullRequest)`

The release still succeeds; only the post-release pom bump automation is incomplete.

**Steps to reproduce:** run `gh workflow run release.yml -f version=v0.1.0` after a successful release pipeline merge; wait for the `bump-version` job; branch push succeeds, PR creation fails (observed on release run 27453957100).

**Expected:** the workflow opens (or updates) a bump PR moving `main` to the next `-SNAPSHOT` version without manual intervention.

**Fix directions (unchanged):** a fine-scoped `GH_PAT`/GitHub App token for the bot with `pull_requests: write`, **or** the repo/org setting allowing Actions to create PRs, **or** downgrade gracefully to an issue+label ping. PR #208 makes the failure loud instead of silent and documents the manual fallback in `RELEASING.md` — merge that first regardless of which fix direction wins.

### Part 2 — version-declaration-driven trigger

Replace the hand-crafted tag with a declared version as the entry point:

- The version lives in one committed file (e.g. `.github/project.yml` `current-version`, or the pom itself as source of truth).
- A PR changing it runs a **pre-release validation** job (version format, changelog presence, milestone state).
- Merging it triggers a job that sets the pom version, creates and pushes the tag, and bumps `main` to the next `-SNAPSHOT` — feeding the existing `verify → scan → promote → release` chain unchanged.
- `workflow_dispatch` with an existing tag stays as the re-release escape hatch it already is.

[...]

PR #575, opened 2026-08-12
fix(review): keep security severity on the defect class, not on uncertainty

### LINKED ISSUE #570: Security severity is inconsistent across equivalent defects, and drifts downward
Measured across the round-2 dogfood corpus (8 languages, 9 planted defects each, answer keys written before the reviews ran).

### The controlled pair

Angular #11 and React #17 plant the same defect: stored XSS through the framework's HTML-injection escape hatch, on user-authored content, with no sanitizer anywhere in the PR.

| | Angular (`bypassSecurityTrustHtml`) | React (`dangerouslySetInnerHTML`) |
|---|---|---|
| severity | HIGH | **LOW** |
| placement | inline comment, top Key Finding | collapsed "Things to double-check" |
| framing | asserted | "confirm the backend sanitizes before storage" |

Same class, same absence of sanitization, opposite handling. The React finding also contradicts itself: its body states the severity was *"deliberately capped at medium"* while it publishes as **LOW**. The PR's risk table then reads `0 Critical / 2 High` for a PR containing stored XSS.

A finding rendered LOW lands in a collapsed section with no inline thread — the surface a reviewer is least likely to read. That is where this one went.

### The broader drift

Independently, scoring of Java #15 and Kotlin #13 found severity deflated by one notch in **ten** instances and inflated in none. Combined with the above, the pattern is a consistent downward bias rather than noise, strongest where the defect requires an assumption about code outside the diff ("maybe the backend sanitizes").

### Why this matters more than a label

Severity drives placement: inline versus collapsed, and whether the check-run holds. Under-rating a security defect does not merely mislabel it — it hides it. And the reasoning that produced the LOW ("the backend might sanitize") is exactly the kind of unverifiable external assumption the prompt's own confidence rules tell it to treat as a verification request rather than a reason to downgrade risk.

### Fix direction

Severity should be a property of the defect class and its blast radius; *confidence* is the right axis for "I cannot see whether something upstream mitigates this". The prompt already separates the two — #545 restated that for three other dimensions — but the security dimension still lets uncertainty pull severity down. Pin equivalent defects to equivalent severity, and let low confidence do the hedging.

A regression test is cheap here: the corpus plants the same XSS class in two frameworks, so the two should score the same severity.

devops-thiago:
> **Stacked on #573** (`fix/569-570-prompt-truth`, issue #569). Both PRs touch `PrReviewPrompts`;
> this one targets that branch so the diff stays reviewable. Merge #573 first, then retarget or
> merge this.

## What type of PR is this?

- [x] 🐛 Bug fix
- [x] 🔒 Security

## Description

The round-2 corpus plants the same defect in Angular and React: stored XSS through the framework's
HTML-injection escape hatch, on user-authored content, with no sanitizer anywhere in the PR.

| | Angular (`bypassSecurityTrustHtml`) | React (`dangerouslySetInnerHTML`) |
|---|---|---|
| severity | HIGH | **LOW** |
| placement | inline comment, top Key Finding | collapsed "Things to double-check" |
| framing | asserted | "confirm the backend sanitizes before storage" |

The React finding also contradicted itself — its body said the severity was *"deliberately capped at
medium"* while it published as **LOW**. Independent scoring across Java #15 and Kotlin #13 found
severity deflated by one notch in ten more instances and inflated in none.

Severity drives placement: a LOW lands in a collapsed section with no inline thread. Under-rating a
security defect does not mislabel it, it hides it — and the reasoning that produced the LOW ("the
backend might sanitize") is precisely the unverifiable external assumption the prompt's own
confidence rules already say to treat as a verification request.

### The fix

Severity is a property of the defect class and its blast radius; *confidence* is the axis for "I
cannot see whether something upstream mitigates this". The prompt separated the two, and #545
restated it for three other dimensions — the security dimension was the one still letting
uncertainty pull severity down. Three changes to `PrReviewPrompts.SYSTEM`, all prompt text:

1. **Dimension 2 gets a severity anchor.** When the provided material shows data a user can author
   reaching an injection sink — the framework HTML escape hatches named together
   (`dangerouslySetInnerHTML`, `bypassSecurityTrustHtml`, `v-html`, `innerHTML`), string-built SQL,
   a shell command or path assembled from input, unsafe deserialization — and no sanitization,
   encoding, escaping or validation is visible anywhere in the material, the finding floors at
   `high`, reaching `critical` when the material also shows the tainted value stored and rendered
   back to other users or the sink reachable unauthenticated. An unshown upstream sanitizer lowers
   *confidence* and must name the exact layer to verify. Explicitly: the same class scores the same
   whatever framework it appears in.
2. **The published risk must match the reasoning.** Added to the existing "Severity is not
   confidence" block: a body that says the severity was capped, held back or hedged cannot publish
   a lower risk field — the hedge goes to confidence, the class goes to risk.
3. **The unshown-artifact cap is scoped, not weakened.** "Claims about … artifacts not shown in the
   diff … are never critical or high" keeps its full force for unshown *artifacts*; it now says it
   does not apply to an unshown *mitigation* for a defect that is shown. That is the reading the
   React finding used to justify LOW on a sink and a tainted value both sitting in the diff.

### What is deliberately not weakened

This round's precision was 7 false positives across 88 findings with every decline trap refused, and
none of the machinery behind that is touched: the self-check list, the omit-rather-than-guess rule,
the quote-verbatim requirements, and every confidence cap keep their current wording. The change
only forbids expressing uncertainty *as a lower severity* — the uncertainty still has to be stated,
in the confidence field and in the description, which is where a reviewer can act on it. Two new
assertions pin that the unshown-artifact cap survives.

## Related Issues

Fixes #570

## How Has This Been Tested?

- [x] Unit tests

Two new `PrReviewPromptsContentTest` cases pin the security severity anchor, the cross-framework
equivalence, the published-risk/stated-reasoning agreement, and — as a regression guard against
over-correcting — that the unshown-artifact cap is still present.

Per the round's standards, a prompt-text change is exempt from red/green; the suite stays green. The
behavioral claim here is about model output, so the real evidence is the corpus: Angular #11 and
React #17 plant the same XSS class in two frameworks and should now score the same severity, which
the next scoring round can measure directly.

### Gates

| Gate | Result |
|---|---|
| `spotless:apply` + `clean compile spotbugs:check spotless:check` | BUILD SUCCESS, `BugInstance size is 0` |
| `clean test` | `Tests run: 2702, Failures: 0, Errors: 0, Skipped: 0` |
| jacoco ∩ `git diff -U0 <parent>...HEAD` (main code) | 0 uncovered lines, 0 uncovered branches |

## Checklist

## Additional Notes

Prompt text only — no `README.md` or `docs/**` file is touched, so the Docs workflow is not in play.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/PrReviewPrompts.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/PrReviewPrompts.java
@@ -50,6 +50,21 @@ public final class PrReviewPrompts {
                and unpinned supply-chain references. A config or hardening weakness visible in the
                diff is as much a finding as a code vulnerability — do not pass over it because it is
                declarative rather than executable.
+               Severity here is a property of the DEFECT CLASS and its blast radius, never of how
+               sure you are about code you were not shown. When the provided material shows data a
+               user can author reaching an injection sink — a framework's HTML-injection escape
+               hatch (dangerouslySetInnerHTML, bypassSecurityTrustHtml, v-html, innerHTML, any
+               raw-HTML render), string-built SQL, a shell command assembled from input, a path
+               joined from input, unsafe deserialization — and NO sanitization, encoding, escaping
+               or validation of that data is visible anywhere in the provided material, the finding
+               is at least "high"; "critical" when the material also shows the tainted value is
+               stored and rendered back to other users, or the sink is reachable without
+               authentication. That some layer you were not shown MIGHT sanitize is a statement
+               about your CONFIDENCE, not about the risk: lower the confidence, name in the
+               description the exact layer to verify, and leave the severity where the defect class
+               puts it. A sanitizer you cannot see is not a sanitizer. Rate the same class the same
+               way whatever framework or language it appears in — the escape hatch with the more
+               alarming name is not the more severe defect.
             3. REGRESSIONS: Does this change break or remove existing behavior? Compare with base commit context.
             4. COMMENT CONTRADICTS CODE: a comment that states something the code it documents does
                NOT do is a defect, not a style note, and it is demonstrable from the diff alone —
@@ -232,6 +247,16 @@ public final class PrReviewPrompts {
               (dimension 8). Each is demonstrable by quoting two lines from the provided material.
               When you have those two lines, report it — do not trade it away against the
               low-severity omission guidance above.
+            - The risk you publish must be the one your own description defends. A finding whose
+              body says the severity was capped, held back, or hedged — "deliberately capped at
+              medium" — while its risk field reads "low" contradicts itself and misplaces the
+              finding. Put the hedge in confidence and the defect class in risk; never let the two
+              disagree.
+            - Equivalent defects get equivalent severity. Before settling on a level, ask what you
+              would give the same defect class in a different framework or language. If the answer
+              differs, the difference is coming from your uncertainty rather than from the defect,
+              and it belongs in confidence — pin the risk to the class and lower the confidence
+              instead.
 
             Confidence calibration:
             - confidence "high" means another reviewer could confirm the issue using only the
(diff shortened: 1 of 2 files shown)


devops-thiago:
The review's low-confidence note — *"Prompt-only fix efficacy relies on model behavior; tests assert text, not output"* — is correct, and worth stating plainly rather than dismissing: the added tests prove the prompt now **says** the right thing, not that the model now **does** the right thing. No unit test can close that gap for a prompt change.

It is verifiable empirically, though, and the instrument already exists. The defect this PR addresses was measured, not guessed: eight PRs in `devops-thiago/ThrillhouseBot-test`, each planting nine defects against an answer key written before any review ran. The specific evidence was a controlled pair — the same stored-XSS class through a framework escape hatch, scored **HIGH inline** on Angular (#11, `bypassSecurityTrustHtml`) and **LOW in the collapsed section** on React (#17, `dangerouslySetInnerHTML`), with the React finding internally contradicting itself ("deliberately capped at medium" while publishing LOW).

So the acceptance test for this change is: re-run those two PRs after merge and compare the severity assigned to the two XSS findings. Equal severity, and inline placement for both, is the pass condition. The corpus keys are at `/Users/thiago/repos/tbtest-answers2/` and the round-2 scoring at `/Users/thiago/repos/tbtest-scores2/`, including the twelve one-notch deflations that motivated the wider part of the change.

No code change from this finding — it names a real limit of the approach rather than a defect in the patch.

Issue #26, opened 2026-06-13
feat: drain in-flight reviews on graceful shutdown


devops-thiago:
### Problem Statement

Reviews run asynchronously on a worker pool. On shutdown (SIGTERM during a rolling deploy or restart) in-flight reviews can be dropped, leaving a check run stuck `in_progress` and a PR without its review.

### Proposed Solution

On graceful shutdown, stop accepting new dispatches, drain in-flight reviews within a bounded timeout, and gate readiness so the orchestrator stops receiving traffic before the worker pool is torn down. Mark any review that cannot finish as failed with a retry hint rather than leaving it hanging.

### Alternatives Considered

- Rely on GitHub re-delivery — unreliable and pairs poorly without idempotency.

### Priority

Important

### Code of Conduct

- [x] I have searched for existing issues and this is not a duplicate.

Issue #25, opened 2026-06-13
feat(github): handle API rate limits with retry and backoff


devops-thiago:
### Problem Statement

A bot that reviews many PRs will hit GitHub primary and secondary rate limits. The GitHub REST/GraphQL clients do not appear to honor `Retry-After` / `X-RateLimit-Reset` or back off, so reviews can fail outright or the app can be flagged for abuse. Only the AI call has retry today (`thrillhousebot.review.ai-retry-base-delay-ms`).

### Proposed Solution

Centralize GitHub calls behind retry with exponential backoff and jitter that honors `Retry-After` and `X-RateLimit-Reset`, caps attempts, and surfaces a clean check-run failure when exhausted. Consider SmallRye Fault Tolerance for the policy.

### Alternatives Considered

- Per-call ad-hoc retries — inconsistent and easy to miss.
- No handling — current state; brittle under load.

### Priority

Important

### Code of Conduct

- [x] I have searched for existing issues and this is not a duplicate.

Issue #24, opened 2026-06-13
feat(dashboard): fan out WebSocket events across replicas


devops-thiago:
### Problem Statement

The live dashboard relies on an in-process broadcaster and replay buffer (`SessionEventBroadcaster`). Under multiple replicas, a client connected to replica A never receives review events produced on replica B, so the live activity feed and streaming-output panel are incomplete.

### Proposed Solution

Publish review events through a shared channel — PostgreSQL `LISTEN/NOTIFY` or Redis pub/sub — and have each replica fan out to its own WebSocket clients. Replicate or document the bounded replay buffer behavior.

### Alternatives Considered

- Sticky sessions — partial; cross-replica events still lost.
- External message broker (Kafka/NATS) — heavier than needed for this signal volume.

### Priority

Important

### Code of Conduct

- [x] I have searched for existing issues and this is not a duplicate.

PR #602, opened 2026-08-12
fix(review): file a second defect as its own finding instead of burying it

### LINKED ISSUE #587: Mock-fidelity defects are analysed correctly but never emitted as findings (2/8)
Measured across the round-3 dogfood corpus (eight languages, nine planted dimensions each, answer
keys written before the reviews ran, scored by independent agents against a fixed brief).

The mock-fidelity dimension scored **2 YES / 6 PARTIAL** across eight languages. "PARTIAL" here has
a very specific and consistent meaning, and it is not "the model missed it":

> the exact reasoning appears, in prose, inside the body of a *different* finding — and is never emitted as a finding of its own.

Per language:

| lang | PR | dim 7 | where the reasoning actually appeared |
|---|---|---|---|
| kotlin | #19 | YES | emitted normally |
| c | #22 | YES | emitted normally |
| scala | #21 | PARTIAL | inside the producer→consumer finding |
| zig | #20 | PARTIAL | a supporting clause inside the pagination finding |
| java | #23 | PARTIAL | a Changed-Files table row |
| csharp | #24 | PARTIAL | noticed three separate times, inside findings 1, 2 and 4 |
| angular | #25 | PARTIAL | inside the comment-inconsistency finding, plus a Changed-Files row |
| react | #26 | PARTIAL | verbatim inside the producer/consumer finding |

Representative — React #26, the sentence sits in another finding's body:

> both fixture scores are below threshold, so it can't distinguish correct filtering from push-everything

That *is* the mock-fidelity defect, stated correctly. It was never filed.

Angular #25 is the starkest: the spec file's own source comment concedes the gap, the review quotes
that concession into a walkthrough row, reuses it as supporting evidence inside finding 5 — and
still never raises it.

### Why this is its own issue, distinct from #537

#537 framed three dimensions as *under-reported* and #545 fixed two of them by rebalancing the
reporting thresholds (round 2: performance 0/4 → 7/8, comment-contradiction 1/4 → 8/8). Mock
fidelity did not respond to that treatment — it went 4/8 in round 2 to 2/8 emitted in round 3, while
the *analysis* is now present in six of eight.

So the remaining failure is not calibration of whether the defect is worth reporting. The model has
already judged it worth writing down. The failure is that a supporting observation made while
building finding A never gets promoted into its own finding B, even when it is a different defect,
in a different file, on a different dimension.

If PARTIALs were counted as caught, round-3 recall would be 63/72 instead of 57/72 — the single
largest recoverable block of recall the corpus has identified.

### Fix direction

Two candidate mechanisms, cheap to distinguish:

1. **A per-dimension emission sweep.** After the findings are drafted, re-read them for statements
   that describe a defect on a dimension not represented in the finding list, and promote each into
   its own finding. The material is already in the response — this is a promotion step, not new
   analysis.
2. **A prompt rule against burying.** State that a defect belonging to a different dimension must be
   filed separately rather than cited as evidence inside another finding, even when it is what makes
   the other finding true.

The corpus is the regression test: it plants exactly one defect per dimension per PR, so a
dimension whose reasoning appears in the response but not in the finding list is mechanically
detectable.

### Related

- #537 — the original dimension-recall measurement (performance and comment-contradiction now fixed
  by #545; mock fidelity is the residue).
- Same round: the performance dimension regressed and failed the same way in three languages, which
  may share this root cause — see the round-3 note on #537.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix

## Description

Mock fidelity scored **2 YES / 6 PARTIAL** across the round-3 corpus, and
"PARTIAL" meant one specific thing every time:

> the exact reasoning appears, in prose, inside the body of a *different*
> finding — and is never emitted as a finding of its own.

React #26 stated the defect verbatim inside the producer/consumer finding
("both fixture scores are below threshold, so it can't distinguish correct
filtering from push-everything"). Angular #25 quoted the spec file's own
concession into a walkthrough row, reused it as supporting evidence inside
finding 5, and still never raised it. C# #24 noticed it three separate times,
inside findings 1, 2 and 4.

So the model has already judged the defect worth writing down — this is not the
reporting-threshold problem that #545's rebalancing fixed for the performance
and comment-contradiction dimensions. The failure is that a supporting
observation made while building finding A never gets promoted into its own
finding B, even when it is a different defect, in a different file, on a
different dimension.

### What changed

`PrReviewPrompts.SYSTEM`, in the block that already governs emission ("Severity
is not confidence, and neither one is a reason to stay silent"), gains the
promotion rule and the sweep:

- Every defect gets its own finding, on the dimension it belongs to. Stating a
  second, different defect inside another finding's description, in a
  `summary.file_summaries` row, or in a `description_gaps` entry is **not**
  reporting it — those surfaces carry no severity, no anchor line and no review
  thread, so a defect that appears only there reaches nobody.
- The rule is explicitly disambiguated from the neighbouring "Report each
  underlying defect exactly once", which forbids restating *one* defect at
  several lines rather than filing a *second* one. Reading the two together is
  a plausible reason the model treats a bury as compliance.
- A closing sweep: re-read each finding description, each `file_summaries` line
  and each `description_gaps` entry for statements describing a defect no
  finding covers, and promote each into its own finding at the risk and
  confidence its own dimension prescribes. Framed as promotion of material
  already written, not new analysis — which is what the issue's option 1 asks
  for and what makes it cheap.

Dimension 8 and the injected `MOCK_FIDELITY_REQUEST` block repeat it at the
point the contradiction is actually noticed: while building the finding it was
evidence for.

## Related Issues

Fixes #587

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

Prompt-text change: exempt from red/green proof per the round's standards, and
covered by the deterministic anchor test the repo already uses for this file.
`PrReviewPromptsContentTest.aDefectOnAnotherDimensionMustBePromotedOutOfTheFindingItSupports`
pins the seven markers (five in `SYSTEM`, one in dimension 8, one in
`MOCK_FIDELITY_REQUEST`) so a future edit cannot silently revert the guidance.

Whether the model *acts* on it is the corpus's job, and the issue states the
regression test it already provides: the corpus plants exactly one defect per
dimension per PR, so a dimension whose reasoning appears in the response but
not in the finding list is mechanically detectable.

### Gates

```
./mvnw -B spotless:apply                                    OK
./mvnw -B clean compile spotbugs:check spotless:check       BugInstance size is 0 / BUILD SUCCESS
./mvnw -B clean test                                        Tests run: 2769, Failures: 0, Errors: 0, Skipped: 0
```

Coverage — jacoco ∩ `git diff -U0 fda4bc7...HEAD`:

```
src/main/java/.../review/ai/PrReviewPrompts.java: 29 changed lines, 0 instrumented,
0 uncovered lines, 0 lines with uncovered branches
RESULT: CLEAN
```

All 29 changed main lines sit inside compile-time `String` constants, so they
carry no bytecode of their own; the class's own initialization is exercised by
the content test.

## Checklist

## Additional Notes

The issue notes that the performance dimension regressed the same way in three
languages and may share this root cause. The rule added here is written per
*defect*, not per dimension, so it covers that case too — and the sweep names
dimensions 4, 5 and 8 as the ones most often lost, since their evidence is
exactly what gets cited in support of something else.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/PrReviewPrompts.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/PrReviewPrompts.java
@@ -151,7 +151,11 @@ public final class PrReviewPrompts {
                nonetheless untested, which is precisely the failure nobody notices — and note that
                the mandated low/medium CONFIDENCE describes how firmly you may word the claim, not
                whether the finding is worth emitting. A green test whose mocks contradict the
-               real collaborator is not evidence the production path works.
+               real collaborator is not evidence the production path works. You will most often
+               notice the unfaithful stub while building some OTHER finding — the fixture that
+               cannot distinguish the two cases, the mock that makes a broken path look proven.
+               Emit the mock-fidelity finding anyway: a contradiction stated only inside another
+               finding's body, or in a walkthrough row, has not been reported.
             9. PRODUCER → CONSUMER CONTRACT: hunks are judged locally, so a change can be correct
                line by line and still wrong end to end. Once per PR, for the change's PRIMARY new
                or modified data structure — a returned collection, a flag, a computed verdict —
@@ -257,6 +261,26 @@ public final class PrReviewPrompts {
               differs, the difference is coming from your uncertainty rather than from the defect,
               and it belongs in confidence — pin the risk to the class and lower the confidence
               instead.
+            - EVERY defect gets its OWN finding, on the dimension it belongs to. While writing one
+              finding you will often state a SECOND, different defect as supporting evidence — a
+              stale comment quoted to show what the code was meant to do, a stub that cannot
+              happen in production, a scan the input does not bound, a page never walked, a
+              fixture that cannot tell the two cases apart. That second defect is a finding in its
+              own right and must be emitted as one. Stating it inside another finding's
+              description, in a summary.file_summaries row, or in a description_gaps entry is NOT
+              reporting it — those surfaces carry no severity, no anchor line and no review
+              thread, so a defect that appears only there reaches nobody. This does not conflict
+              with "report each underlying defect exactly once" below: that rule forbids restating
+              ONE defect at several lines; this one forbids burying a SECOND defect inside the
+              first, even when it is what makes the first one true.
+            - Before you finish, re-read what you have written — each finding's description, each
+              file_summaries line, each description_gaps entry — for any statement that describes
+              a defect no finding in your list covers, and promote each one into its own finding
+              at the risk and confidence its own dimension prescribes. The material is already
+              written, so this is a promotion step, not new analysis. The dimensions this loses
+              most often are the ones whose evidence is naturally cited in support of something
+              else: an unfaithful stub (dimension 8), an added quadratic shape (dimension 5), and
+              a comment contradicting the code (dimension 4).
 
             Confidence calibration:
             - confidence "high" means another reviewer could confirm the issue using only the
@@ -790,6 +814,10 @@ public final class PrReviewPrompts {
               declared SIGNATURE counts as shown material: a stub that returns null against a
               non-null contract, throws a checked exception the method does not declare, or
               returns a value the declared type excludes is contradicted by the signature alone.
+            - File it as its own finding. Noticing the contradiction while building a different
+              finding is the usual case — it turns up as the reason that finding's evidence is
+              weak — and citing it there, or in a walkthrough row, does not report it. Emit the
+              mock-fidelity finding in addition to the one you were writing.
             - A faithful stub that matches the real contract is not a finding."""
           .stripIndent();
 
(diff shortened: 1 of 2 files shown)


devops-thiago:
Addressing the open low-confidence finding, since it sits in the summary section with no inline thread and would otherwise read as unhandled.

> **LOW:** Fix efficacy unverifiable from this diff: test pins prompt markers, not model behavior (`PrReviewPrompts.java:264`)

**Accurate, and deliberately not fixed in code.** `PrReviewPromptsContentTest` asserts that the promotion-rule text is present, which is the most a prompt diff can prove — behaviour can only be demonstrated by running the model. Prompt-text changes are exempt from the red/green requirement for exactly this reason, so there is no test to add here that would answer the finding.

The caveat itself stands and is worth recording rather than dismissing: **this PR's effect on #587 is unverified.** Mock-fidelity emission currently measures 2/8 across the round-3 corpus (the reasoning appears inside another finding's body in six of eight languages and is never filed as its own finding). Nothing in this diff demonstrates that number moves.

The proof is a corpus re-run against a deployed image containing this change, scored against the existing answer keys — the same method that confirmed #573 at 8/8 and, importantly, caught #575 still failing after it looked correct on paper. Dimension 7 rising above 2/8 is the acceptance test.

Two sibling changes in this batch are in the same position and will be measured in the same run: #594 (equivalent injection sinks must publish equivalent severity — #570's acceptance test, which #575 failed once already) and #596 (one claim on one surface).

No code change on this PR.

Issue #508, opened 2026-08-10
Valid review JSON with a mis-shaped summary field fails the whole review — retried 5× at full price


devops-thiago:
## Summary

Production log analysis (v0.5.0, 2026-08-09/10): the reviews of PR #504 (session 2309) and PR #495 (session 2302) both failed all 5 attempts with `Model response is not valid review JSON`. The response was **not** invalid JSON. Reconstructed from the logged SSE stream:

- `finish_reason: "stop"` — no truncation anywhere in the log (zero `length` stops in 24 completions)
- the body parses cleanly: `{findings, previous_findings_status, summary}`
- Jackson mapping then fails:

```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException:
  Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)
  at ... StringCollectionDeserializer.deserialize ...
  at ReviewResponseParser.parse(ReviewResponseParser.java:44)
```

The culprit field, from the reconstructed response: **`summary.description_gaps`** — schema expects `List<String>`, the model (deepseek-v4-flash, reasoning effort=high) emitted objects:

```json
"description_gaps": [{"claim": "Additional Notes: 'no max_tokens is sent…'", "code": "…"}]
```

`findings` and `previous_findings_status` mapped fine. One mis-shaped *summary* field discarded the entire response.

## Why this is the #492 family, not a one-off

`ReviewResponseParser.parse` reports every `IOException` — including a Jackson **mapping** failure on structurally valid JSON — as "not valid review JSON", which routes it into the transient-failure retry lane. But a schema-shape failure is (near-)deterministic for a given model+prompt: session 2309 failed the same way **5/5 at temperature 1.0**. Cost per attempt ~52,900 tokens (~29,900 in + ~23,000 out, of which **93% reasoning tokens**), so one PR's failed review burned ≈265K tokens producing nothing — the same shape as the truncation retry storm #492/#495 removed, through a different door.

The parser already solves exactly this class for one field: `normalizePreviousFindingsStatus` exists because "models sometimes emit `previous_findings_status` as an object … instead of the array the schema asks for. Normalize it … so the shape mismatch does not fail the whole review (and force a full-cost retry)." `description_gaps` needs the same tolerance.

## What a fix needs to cover

- **Normalize `description_gaps`** objects → strings (join the object's string values, or take a designated field), following the existing `normalizePreviousFindingsStatus` pattern.
[...]

PR #688, opened 2026-08-13
fix(ai): fail fast on a provider context-length rejection instead of retrying it at full price

### LINKED ISSUE #622: A provider context-length rejection is retried 5x at full price
Split out of #562 so it is not lost when #619 closes that issue. #619 fixes the shipped defaults and adds startup rules; this is item 3 of the original, deliberately left out because it lives in the AI call/retry lane rather than in config.

When the provider rejects a request because it exceeds the context window, the review retries the call **five times**, paying full price for each, before failing. A context-length rejection is deterministic: the same prompt against the same window fails identically every time. Retrying it cannot succeed and only multiplies the cost of a request that was already too large.

Same class as #495 and #508 — a non-retryable provider response treated as a transient failure.

### Why it matters beyond the money

The failure mode compounds with the problem that produced it. A review that overruns the window is, by construction, one carrying a large diff or a large accumulated context (see #583, where a single request reached 1,768,374 characters). So the requests most likely to be rejected are also the most expensive ones to retry, and the maintainer waits through five of them before learning the review will not complete.

### Suggested direction

Classify a context-length rejection as terminal rather than transient, fail fast, and surface it as a disclosure the maintainer can act on ("the diff exceeds the model's context window") rather than a generic failure after five attempts. If a retry is worth attempting at all it should be a *different* request — a smaller batch — not the same one again.

### Where to look

The retry policy around the AI call, alongside the existing truncation and error classification. #619's author identified this but scoped it out deliberately: bundling a retry-policy change into a config correction would have made both harder to review and to revert.

Not a release blocker on its own — it costs money and time on requests that were already failing, rather than producing wrong output.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix

## Description

A provider rejection for exceeding the model's context window is deterministic — the identical request against the identical window is rejected identically — yet the AI call retry lane treated it as transient and retried it up to `max-ai-retries` times, billing every attempt in full. The requests most likely to be rejected are, by construction, the largest (and so most expensive) ones a review makes.

This PR classifies the rejection as terminal, fails fast, and surfaces it as a disclosure the maintainer can act on, following the same handling shape as the #495 truncation and the #508 spend-ceiling refusal:

- **New `AiContextWindowExceededException`** (extends `AiReviewException` so it survives the executor unwrap with its identity intact), with a message-based classifier covering the wordings OpenAI (`context_length_exceeded`, "maximum context length"), Anthropic ("prompt is too long", "exceed context limit"), and Gemini ("exceeds the maximum number of tokens") use for the input-side rejection. A marker miss degrades safely to today's retry-as-transient behaviour.
- **`AiReviewService.runWithRetries`** classifies each failure before queueing a retry: a context-window rejection is thrown immediately after the first attempt — no retries, no retry events on the dashboard.
- **Multi-call lane (`FindingPipeline`)**: a rejected batch goes straight to disclosure (its files recorded as not reviewed) instead of the sequential retry that would re-send the identical prompt; sibling batches keep their findings.
- **Single-call lane (`ReviewOrchestrator` / `ReviewPublisher`)**: the failure notice and the FAILED check run name the deterministic cause and point at `REVIEW_MAX_INPUT_TOKENS` (whose token budgeting batches or clips the diff to fit), instead of the generic "reply with `/review`" advice that would repeat a knowably futile request.

## Related Issues

Closes #622

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

`./mvnw verify` passes (Spotless, SpotBugs, unit + integration test suites). New tests cover the classifier's provider wordings, case-insensitive cause-chain walking and cyclic-chain bound; the retry lane making exactly one call and broadcasting no retry events; the multi-call batch disclosure with sibling findings kept; and the orchestrator's notice/check-run copy including the buried-in-cause-chain case. Patch coverage on changed lines verified locally against a fresh JaCoCo report from the merge-base diff.

## Checklist

## Screenshots / Logs

N/A

## Additional Notes

Changelog updated under `[Unreleased]`. Not a release blocker on its own — the bug costs money and time on requests that were already failing, rather than producing wrong output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to ThrillhouseBot.
 
 ### Fixed
 
+- **A provider context-length rejection is no longer retried at full price** (#622): a rejection for exceeding the model's context window is deterministic, so the review call now fails fast on the first attempt instead of re-billing up to `max-ai-retries` identical requests. In a multi-call review the rejected batch's files are disclosed as not reviewed while the other batches keep their findings; a single-call review fails with a notice and check run that name the cause and the `REVIEW_MAX_INPUT_TOKENS` knob to lower, instead of generic retry advice
 - **Dashboard token test no longer assumes an en-US locale** (#661): the test now asserts the same `toLocaleString()` output the component renders, so it passes on machines with any runtime locale
 
 ## [0.6.0] — 2026-08-13
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingPipeline.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingPipeline.java
@@ -21,6 +21,7 @@ import dev.thiagogonzaga.thrillhousebot.config.BotIdentity;
 import dev.thiagogonzaga.thrillhousebot.dashboard.ReviewSession;
 import dev.thiagogonzaga.thrillhousebot.github.GitHubPullRequestClient;
 import dev.thiagogonzaga.thrillhousebot.github.GitHubReviewClient;
+import dev.thiagogonzaga.thrillhousebot.review.ai.AiContextWindowExceededException;
 import dev.thiagogonzaga.thrillhousebot.review.ai.AiResponseTruncatedException;
 import dev.thiagogonzaga.thrillhousebot.review.ai.AiReviewService;
 import dev.thiagogonzaga.thrillhousebot.review.ai.FindingVerificationService;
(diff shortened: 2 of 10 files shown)


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/AiContextWindowExceededException.java:105 — Classifier substring markers can over-match non-input or transient provider failures

Issue #628, opened 2026-08-12
Patchless files are disclosed as having exceeded the review budget


devops-thiago:
Split out of the #532 finding triage (PR #627), which verified the defect as real but deliberately left it unfixed — the honest fix is disproportionate to the symptom and touches release-critical plumbing.

### The defect

A file with no patch (binary, or a rename GitHub returns without content) is disclosed to the maintainer as though it were dropped for exceeding the review budget. It was not: there was nothing to review in the first place. The wording tells the reader the review ran out of room on a file that never had content, which points them at raising a budget that would change nothing.

Severity is genuinely LOW — it is a wording defect on a disclosure, not wrong analysis.

### Why it was deferred rather than fixed

Distinguishing the two cases at the point of disclosure means carrying the reason a file was omitted, which requires adding a component to both `BudgetPlan` and `TruncationDetail`. Those are records with canonical constructors called from **57 and 29 sites** respectively, all on the budgeting and disclosure path that the round-4 work has just stabilised.

Trading that much churn on release-critical plumbing for a LOW wording fix, immediately before a release, is the wrong trade. Recorded here so it is not lost.

### Suggested direction when it is picked up

Carry an omission *reason* rather than inferring it from which list a filename lands in. The lists already exist (`omittedFiles`, `clippedFiles`, ceiling-skipped); what is missing is why. Doing it as a deliberate small refactor — with the call sites updated in one pass rather than opportunistically — is much safer than threading a boolean through.

Worth pairing with the related disclosure work if any of it is revisited: #515 (ceiling-skipped files keeping walkthrough rows) and the ceiling-skipped labelling fixed in #627 are the same family — the reader being told a file was skipped without being told accurately why.

PR #691, opened 2026-08-13
fix(review): failed-call coverage clause and legacy omitted count alongside summary degradations

### LINKED ISSUE #655: Coverage banner blames the diff budget for files whose review call failed
The coverage banner gives one reason for three different causes, and for a failed review call the
reason it gives is wrong.

When a batch fails all its retries, `FindingPipeline` calls `plan.recordUncoveredFiles(...)`, which
lands in `BudgetPlan.runtimeUncoveredFiles()`. `BudgetPlan.effectiveOmittedFiles()` merges that list
into the planned omissions, and `VerdictBuilder` builds `TruncationDetail.omittedFileNames` from
`withoutNames(plan.effectiveOmittedFiles(), ceilingSkipped)`. `ReviewResult.coverageGapClause` then
renders every name in it as withheld

> because the diff exceeded the review budget

The same review's summary overview separately says of those files

> (not reviewed — the review call for it did not complete; treated as uncovered)

so the two surfaces give contradictory reasons for the same file, and the banner's implied remedy —
raise the input budget — does not apply to a call that failed.

The spend-ceiling class already carries its own reason (#524, #530) and the response-cut class
carries its own (#523). Runtime call failure needs the same treatment: either its own clause, or a
clause that does not name budget exhaustion when that is not the cause.

Related: #628 covers the same wrong clause reached by a different route, a file GitHub reported with
changes but no patch text. A fix that gives each cause its own reason would close both.

Surfaced by the bot's review of the v0.6.0 release PR (#532).

### LINKED ISSUE #659: Coverage clause drops the legacy omitted-file count when the summary also degraded
`ReviewResult.coverageGapClause` drops the legacy omitted-file count whenever the same review also
has a summary degradation to disclose. The count is not merely misworded, it disappears.

The method opens with the fallback:

```java
if (detail == null || detail.isEmpty()) {
  return omittedFilesClause(omittedFiles);
}
```

`TruncationDetail.isEmpty()` is `!hasFileGaps() && summaryDegradation == NONE`. A detail carrying
only a summary degradation is therefore *not* empty, so the fallback is skipped — and nothing below
it ever reads the `omittedFiles` int again. Every remaining clause is built from the four name lists,
which are empty on the legacy path.

Probed directly against the current `main`:

```
A  legacy count, empty detail   -> 3 file(s) were omitted because the diff exceeded the size budget
B  legacy count + summary cut   -> the summary was shortened because the model's response was cut at
                                   its length cap (max-output-tokens / REVIEW_CONCISE_MAX_OUTPUT_TOKENS)
                                   — the findings themselves are complete
C  legacy count + summary skip  -> the summary was skipped because the review's token spend ceiling
                                   (REVIEW_MAX_TOKENS_PER_REVIEW) was reached — the findings
                                   themselves are complete
```

In B and C the reader is told the summary was shortened and never told that three files went
unreviewed. That is the coverage disclosure this class exists to guarantee.

The javadoc above the method states the fallback as unconditional — "or the legacy numeric clause
when no names are known" — which is true only when the detail is empty.

### Reachability

Both halves are needed at once: `omittedFiles > 0` with no names (the legacy line cap, so
`max-input-tokens=0`, not the default) plus a degraded summary. The summary call can be cut at its
length cap under any budgeting setting, so the combination is reachable on a deployment that has
turned token budgeting off.

### Direction

The numeric fallback belongs with the other per-class clauses rather than as an early return: when no
names are known but `omittedFiles > 0`, add the legacy clause to `clauses` and let the degradation
clause join it.

### How this was found

The bot reported it on the release PR (#532) and its own verifier rejected the finding, correctly, on
the strong claims it made: that `coverageGapClause` never reads `omittedFiles`, and that the legacy
path renders an *empty* clause. Neither is true — the fallback does read it on the empty-detail path,
and the clause is never empty. The narrower defect above survives those corrections, and was lost
with the overstatement. Same shape as #650.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix

## Description

Two defects in `ReviewResult.coverageGapClause`, both about the coverage disclosure telling the reader the wrong thing (or nothing) about unreviewed files:

- **#655 — wrong reason for failed review calls.** Files a failed batch left unreviewed flowed `recordUncoveredFiles → runtimeUncoveredFiles → effectiveOmittedFiles → TruncationDetail.omittedFileNames`, so the banner said they were withheld "because the diff exceeded the review budget" while the summary overview said "the review call for it did not complete" — contradictory reasons, and the banner's implied remedy (raise the input budget) cannot help. Runtime call failures are now their own `TruncationDetail` class (`callFailedFileNames`), carried explicitly the same way the spend-ceiling (#524/#530) and response-cut (#523) classes are, and disclosed as "not reviewed because the review call for them did not complete" in the coverage clause and the check-run brief. `VerdictBuilder` subtracts ceiling skips before classifying, so a deliberate ceiling stop is never relabelled a failure, and approval gating / omitted counts are unchanged.
- **#659 — legacy count dropped next to a summary degradation.** On the legacy line-cap path the numeric fallback was an early return guarded by `detail.isEmpty()`; a detail carrying only a summary degradation skipped it, and nothing below read the `omittedFiles` int, so the count vanished (issue probes B/C). The numeric clause now joins the per-class clauses when no names are known and `omittedFiles > 0`, so it renders alongside the summary-degradation clause. Probe A (empty detail) is unchanged.

The pre-existing `TruncationDetail` shape is kept via a convenience constructor, leaving existing call sites (including #628's future patchless-reason work) untouched. #628's patchless route is deliberately out of scope here.

## Related Issues

Closes #655
Closes #659

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

New tests cover the failed-call class end-to-end from `BudgetPlan.recordUncoveredFiles` through `VerdictBuilder` to the rendered banner and check-run brief (including no-double-listing with ceiling skips and clipped files), plus the issue's A/B/C probes for the legacy count. `./mvnw verify` passes locally (3133 tests) and every changed production line is covered against the merge-base diff.

## Checklist

## Screenshots / Logs

N/A

## Additional Notes

CHANGELOG `[Unreleased]` carries Fixed entries for both issues.


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewResult.java:564 — Legacy omitted count still dropped when a per-file class coexists (e.g. call failure)

The combination cannot be produced: `coverageGapClause` has exactly three producers, and none can pass a detail whose `callFailedFileNames` is non-empty while `omittedFiles` exceeds the named total.

- On the legacy line-cap path (`plan.budgeted() == false`), `VerdictBuilder.build` assigns `truncation = ReviewResult.TruncationDetail.EMPTY` unconditionally — a failed batch's `runtimeUncoveredFiles` never reach the detail there, so `hasFileGaps()` is false and the numeric clause renders.
- On the budgeted path, the count is computed as `plan.effectiveOmittedFiles().size() + clipped.size() + responseCut.size()`, and the detail's name classes are a partition of that exact total (`omittedFileNames` = effective omissions minus ceiling skips minus call failures, plus `spendCeilingSkippedFileNames`, `callFailedFileNames`, `clippedFileNames`, `responseCutFileNames`) — the count can never exceed the sum of the named classes, so the numeric clause has no residue to disclose.
- The remaining producers (`AbstractPrSuggestionGenerator.disclosure`, `FollowUpDeltaSummary`) pass a count equal to their own name lists and never populate `callFailedFileNames`.

Gating on "count exceeds the named total" would guard a state no producer can construct.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Issue #680, opened 2026-08-13
fix(review): reconcile dimension 7's mismatch rule with the unshown-producer escape


devops-thiago:
## Background
In the review system prompt (`PrReviewPrompts`, dimension 7), the rule "treat a mismatch as high" is defined against "NOTHING IN THE PROVIDED MATERIAL PRODUCES …", while the unshown-state escape for externally produced artifacts (base image, release binary) sits two sentences below, unreconciled with that definition. A model reading the definition literally can floor a mismatch on an artifact whose producer is legitimately outside the diff, exactly the case the escape was written for. Raised on PR #641 (thread `r3769584139`, still unresolved); hedged and model-behavioral, but the textual tension is real and unchanged on main.

## Change required
In `src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/PrReviewPrompts.java`, dimension 7: fold the external-producer carve-out into the mismatch definition itself (or cross-reference it from the definition sentence), so the two rules cannot be read in isolation as contradicting.

## Constraints
- Prompt-only change; no behavior change elsewhere. Keep the dimension's intent: genuine unproduced-artifact mismatches still rate high.

## Acceptance criteria
- [ ] The mismatch definition and the unshown-producer escape read as one rule.
- [ ] Prompt-rendering tests (including `AiServicePromptRenderingTest` coverage of the service) stay green.
- [ ] `./mvnw verify` clean.

## References
- PR #641 review thread: `r3769584139`
- Current implementation: `src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/PrReviewPrompts.java` (dimension 7)

Issue #679, opened 2026-08-13
fix(review): CLEAR_DIRECTIVE hardcodes @thrillhousebot instead of the configured bot login


devops-thiago:
## Background
`FollowUpAnalyzer.CLEAR_DIRECTIVE` is `Pattern.compile("@thrillhousebot\\s+resolved\\b(?!\\s*\\?)")` — the bot's login is a hardcoded literal. The rest of the codebase resolves the bot's identity from the configured bot-logins setting, so on an install whose GitHub App uses a different login the clear-directive feature is silently dead: maintainers mention their bot's real name and nothing matches. Raised twice on the v0.6.0 release PR and not addressed (#647 reworked the acks and locator guards, not the literal).

## Change required
In `src/main/java/dev/thiagogonzaga/thrillhousebot/review/FollowUpAnalyzer.java`: build the directive pattern from the configured bot login(s) (quote them with `Pattern.quote`), keeping the `resolved\b(?!\s*\?)` interrogative guard intact. Compile once per config, not per comment.

## Constraints
- Default-config behavior must be unchanged (`@thrillhousebot resolved` still matches; `resolved?` still refused).
- Case-insensitivity preserved.

## Acceptance criteria
- [ ] A custom-login install's mention (e.g. `@my-review-bot resolved`) triggers the clear directive.
- [ ] `@thrillhousebot resolved` and the interrogative rejection behave exactly as today under default config.
- [ ] Unit test covers a non-default login.
- [ ] `./mvnw verify` clean.

## References
- PR #532 review threads: `r3762054038`, `r3764530534`
- Current implementation: `src/main/java/dev/thiagogonzaga/thrillhousebot/review/FollowUpAnalyzer.java` (`CLEAR_DIRECTIVE`, ~line 757)

Issue #678, opened 2026-08-13
perf(review): SummarySurfaceDeduplicator re-derives phrase sets per claim pair


devops-thiago:
## Background
`SummarySurfaceDeduplicator.sharesPhrase` calls `phrases(claim)` inside the per-candidate loop, re-deriving the claim's phrase set (and joining strings) once per (candidate, claim) pair; `Claim` stores only `List<String> words`. Flagged on the v0.6.0 release PR (thread `r3768292668`) and left unfixed. Practical risk is low — the input is one model summary — but it is the same quadratic-rederivation shape #647 removed from the recorders, and the fix is local.

## Change required
In `src/main/java/dev/thiagogonzaga/thrillhousebot/review/SummarySurfaceDeduplicator.java`: derive each claim's phrase set and word set once (e.g. carry them on `Claim`, or precompute maps keyed by claim) and reuse them across comparisons.

## Constraints
- Deduplication decisions must be byte-identical: same collapses, same contradiction holds (polarity logic untouched).

## Acceptance criteria
- [ ] `phrases(...)`/word-set derivation happens once per claim, not per pair.
- [ ] Existing deduplicator tests green, no behavioral change.
- [ ] `./mvnw verify` clean.

## References
- PR #532 review thread: `r3768292668`
- Current implementation: `src/main/java/dev/thiagogonzaga/thrillhousebot/review/SummarySurfaceDeduplicator.java` (`sharesPhrase`, `phrases`, `Claim`)

Issue #676, opened 2026-08-13
fix(review): close three under-fire gaps in the injection-sink floor's defeaters


devops-thiago:
## Background
The injection-sink floor (`FindingVerificationService`, #570/#594/#608/#610/#614) floors an unmitigated-injection-sink finding at HIGH. Three defeater-pattern gaps make it under-fire — the floor stays silent on genuine findings. All three were raised in merged-PR review threads and never landed:

1. **`MITIGATION_ABSENT` cannot match "Nothing sanitizes …"** — its negator group is `(no|not|without|missing|never|lacks?|absent)`; `nothing`/`nobody` are absent, so "Nothing sanitizes the value before innerHTML" never registers as an absence claim and the floor cannot fire. Notably, `nothing` **was** added to `MITIGATION_ASSERTED`'s negative lookahead, so the word was on the radar — only the absence side was never widened. (PR #594 thread `r3763526759`.)
2. **`SINK_DENIED` swallows "does not prevent SQL injection"** — the pattern is `\b(no|not|never)\s+(\w+[\s-]+)?` + sink alternation; the one-word gap admits bridge verbs like *prevent/stop/block*, so an assertion of the defect is read as a denial of the sink and suppresses the floor. (PR #594 thread `r3763526798`.)
3. **`CONDITIONAL_CLAUSE` truncates at a coordinator inside the protasis** — the span's negative lookahead stops at `but|so|then|however|therefore|otherwise` anywhere, so "If, however, the API always sanitizes the body on write, …" ends the hypothetical span at "however" and leaves "sanitizes" in text read as asserted, defeating the floor. The PR #610 reply cited fix commit `2697ced`, but that SHA does not exist in the repository — the fix never landed, and no test places a coordinator inside a protasis. (PR #610 thread `r3766530145`.)

## Change required
In `src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/FindingVerificationService.java`:
- Add `nothing|nobody` to `MITIGATION_ABSENT`'s negator alternation.
- Exclude bridge verbs (`prevent|stop|block|guard|protect` forms) from `SINK_DENIED`'s one-word gap (negative lookahead), keeping "no SQL injection here" a denial.
- Let `CONDITIONAL_CLAUSE`'s terminator lookahead ignore a coordinator that is still inside the protasis (e.g. require the coordinator to follow a comma-closed protasis, or match parenthetical coordinators like ", however," as non-terminating), and document the residual trade-off.

[...]

PR #695, opened 2026-08-13
chore(release): prepare 0.6.1

### LINKED ISSUE #622: A provider context-length rejection is retried 5x at full price

devops-thiago:
## What type of PR is this?

- [x] 📝 Documentation

## Description

Prepares the 0.6.1 release: bumps the project version from `0.6.1-SNAPSHOT` to `0.6.1` and turns the `[Unreleased]` changelog section into `[0.6.1] — 2026-08-13`, with a short intro for someone running 0.6.0.

All 15 fixes address behavior that shipped in 0.6.0, so every entry is kept. The Dependencies section carries the docs and frontend group updates (#658, #662, #689) plus the typescript 6.0.3 → 7.0.2 bump (#369), which had no changelog entry until now. Checked before rolling: no open Dependabot PRs, no open security alerts, and the v0.6.1 milestone is closed.

The website version dropdown still lists v0.6.0 as current; archiving docs for the dropdown has been done in separate bulk commits before (see 4acf223), so it is left out of this PR.

## Related Issues

N/A — release preparation. The fixes themselves closed #622, #623, #626, #636, #652, #653, #654, #655, #656, #659, #661, #676, #678, #679, #680.

## How Has This Been Tested?

- [x] Manual testing

Version and changelog edits only; no code changes. CI runs the full suite on this PR.

## Checklist


devops-thiago:
@thrillhousebot resolved CHANGELOG.md:28 — Changelog names REVIEW_MAX_INPUT_TOKENS without stating its default

Issue #755, opened 2026-08-16
Log forgery through a model-supplied finding path survives MarkdownSafe.oneLine, so #742 is only half fixed


devops-thiago:
Found by the audit of the v0.6.5 candidate (`v0.6.4..73cc33d`). #744 closed #742, but the tool it used does not cover the threat #742 named.

## The defect

`MarkdownSafe.oneLine` collapses with `Pattern.compile("\\s+")` — `MarkdownSafe.java:43`. That is the ASCII six: space, tab, LF, VT, FF, CR.

`GitHubApiError` faced this exact threat and fixed it in #740 with a deliberately wider class — `[\s\p{IsCc}\p{IsCf}

]+` at `GitHubApiError.java:191`, whose javadoc at `:167-187` names the survivors of `\s+` by hand and explains why each one matters to a log reader. That reasoning never crossed over to the sanitizer #744 chose.

So a model-supplied `finding.file()` carrying **NEL (U+0085), U+2028, U+2029, NUL, ESC or a bidi override (RLO)** reaches the warn lines at `ReviewPublisher.java:770` and `:832` intact. `String.strip()` at `MarkdownSafe.java:57` does not catch them either — it trims ends, and these are interior.

A log viewer, terminal or JSON/ECS shipper may treat any of them as a record boundary or a screen-control sequence, which is the forged-record vector #742 was filed about.

## Proof

`AuditLogForgeryProbeTest` drives `postInlineComments` through both real `Log.warnf` sites and captures the emitted `LogRecord`. Verbatim failure at `73cc33d`:

```
U+0085 survived MarkdownSafe.oneLine into the warn line:
GitHub rejected inline comment for app.js<NEL>2026-08-16 12:00:00 WARN [thrillhousebot] approved the pull
request, 0 findings<U+2028>…<U+2029>…<NUL>…<ESC>[2K…<RLO>desrever:2 (status=422 …) ==> expected: <false> but was: <true>
```

## Second half: the sweep #742 asked for was not done

#744 changed two WARN lines. Six further sites still interpolate raw model-supplied `title` / `file` into log lines at **INFO**:

- `FindingQuoteValidator.java:83`
- `FindingVerificationService.java:1229`
- `FrameworkFalsePositiveFilter.java:74`
- `FindingPipeline.java:1370`
- `FindingDeduplicator.java:55`
- `FollowUpAnalyzer.java:633`

[...]

Issue #746, opened 2026-08-16
Credential shapes: the bound strands a token prefix, and the JWT shape masks ordinary text


devops-thiago:
Found by deep-audit round 7 against the shipped v0.6.4. Both introduced by #740 (rework) — two halves of the same oversight in `CREDENTIAL_SHAPED_PREFIX` / `CREDENTIAL_SHAPED_VALUE`.

**1. A cut token strands its prefix unmasked (security, bounded).** #740 widened the JWT alternative so a token the 1024-char pre-cut split is still masked, but the same reasoning was never applied to the other three shapes. `ghp_\w{10,}`, `github_pat_\w{10,}` and `bearer\s+[\w.~+/=-]{10,}` still require ten characters, so when the bound severs one below that floor the match fails and **up to nine characters of the secret reach the warn line**. v0.6.3 redacted the whole body before capping and masked the same input. Red proof:

```
Audit7BProbeTest.doesNotStrandACredentialPrefixTheBoundCutBelowTheQuantifier
***,,,…,,,ghp_A1b2C3d4E…  expected: <false> but was: <true>
```

Bounded at ≤9 characters — the cap site cuts after redaction and can never strand. Fix direction: lower the three floors to `{4,}`; the sigil is the discriminator, not the length.

**2. The widened JWT shape masks ordinary text (over-masking).** `eyJ[\w-]{8,}(?:\.[\w-]*){1,2}` compiles under `(?i)`, so it matches `eyj` in any case followed anywhere by a dot. Measured: `eyjafjallajokull.internal.example.com` → `***.com`, a 400-char base64 blob → `***`, and `eyJhbGciOiJIUzI1NiJ9-report-…final.json` → `***`. That is precisely the outcome the shape's own javadoc says it was narrowed to avoid — and worse, it can swallow throttle wording, turning `isThrottled()` false on a body that GitHub meant as a throttle. Fix direction: scope `(?i)` to the bearer half (a JWT header is always literally `eyJ`) and anchor with `(?<[\w-])`.

Doc nit for the same change: `\p{IsCf}` is in the shipped collapse pattern but missing from the javadoc that exists to explain that class.

Proof and refutations: audit report AUDIT7-B (findings B1, B3, B5).

Issue #742, opened 2026-08-16
Model-supplied finding paths reach warn lines uncollapsed, so a crafted path can forge a log record


devops-thiago:
Found by deep-audit round 6 (AUDIT6-A finding A4, second half). It was scoped out of #740 because that PR owned `GitHubApiError` while these two call sites live in `ReviewPublisher`, and it was never filed on its own — this issue closes that gap.

The error-body path now collapses whitespace, C0/C1 controls and Cf format controls before anything reaches a log line (#740). The finding path does not: `finding.file()` is model-supplied text interpolated straight into two warn-level lines in `ReviewPublisher` (around `:752` and `:814`), the second of which #723 promoted from debug to warn — so it is written to production logs today.

A path carrying a line terminator splits the record and forges a second one; a path carrying bidi or other format controls reorders what an operator reads. That is the same log-forgery family #740 closed on the sibling surface, and it undercuts the diagnostic value #723 was added to provide: the whole point of those lines is that an operator can trust what a rejection says.

Fix direction: route both interpolations through the same one-line collapsing the error body uses (`MarkdownSafe.oneLine` or the shared collapse), rather than trusting the model's path string. Worth a sweep for other model-supplied values interpolated into log lines while in there.

PR #758, opened 2026-08-16
fix(github): take the credential value floors to a single character

### LINKED ISSUE #757: The credential-shape floors still strand up to three characters of a token the bound cut, so #746 is PARTIAL
Found by the audit of the v0.6.5 candidate (`v0.6.4..73cc33d`). Introduced by `df4d727` (#750).

## The defect

#746 lowered the value floors from `{10,}` to `{4,}` on all three shapes — `GitHubApiError.java:93` for `gh[pousr]_` and `github_pat_`, `:138` for `bearer` — because `clean()` cuts the collapsed body to 1024 characters *before* `redactCredentials`, and a token severed below the floor stops matching and reaches the log.

That shrinks the leak from at most nine characters of the secret to at most three. It does not close it. A cut landing so that 1–3 value characters remain on the visible side still falls below `{4,}`, and the prefix reaches the warn line unmasked — the same mechanism, one third the size.

The reason this is worth an issue rather than a shrug is the second half: `GitHubApiError.java:87-89` and #750's own body both describe this case as **closed**. The code under-delivers against its own documentation, which is what stops the next reader from checking.

## Proof

`Audit8AFindingTest`, reusing #746's own fixture. Verbatim failure at `73cc33d`:

```
org.opentest4j.AssertionFailedError: ***,,,…,,,ghp_A1b… ==> expected: <false> but was: <true>
org.opentest4j.AssertionFailedError: ***,,,…,,,Bearer S3c… ==> expected: <false> but was: <true>
org.opentest4j.AssertionFailedError: ***,,,…,,,github_pat_A1b… ==> expected: <false> but was: <true>
```

## Refutation attempts

- *"Three characters aren't a secret."* Partly survives — hence the low severity. What does not survive is the documentation claiming the case is handled.
- *"`{4,}` is a deliberate trade."* Refuted. Patching all three shapes to `{1,}` gives **96 tests, 0 failures** with every over-masking control still green (`gh_2.40.0`, `ghost_writer`, `ghp_ prefix` unchanged), and **no test pins `{4,}`**. It was an arbitrary stopping point, not a reasoned floor.

## Suggested fix

`{1,}` on all three shapes. That is what the javadoc's own argument — "the sigil is the whole discriminator" — implies; `ghp_`, `github_pat_` and `Bearer ` do not occur in prose, so the length was never carrying the discrimination.

## Severity

Latent, low. Same precondition as #746 itself: a non-GitHub host returning a large error body — a GHES or reverse-proxy error page, a misconfigured base URL, a compromised endpoint.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [ ] ✨ Feature
- [ ] 📝 Documentation
- [ ] 🔧 Refactor
- [ ] 🚀 Performance
- [ ] ✅ Test
- [x] 🔒 Security
- [ ] 📦 Dependency update
- [ ] 🏗️ CI/CD

## Description

`clean()` cuts a response body to 1024 characters *before* `redactCredentials` runs, so a
credential the cut severs can fall below its shape's length floor, stop matching, and reach the
warn line with whatever survived the cut intact.

#746 lowered the three floors from `{10,}` to `{4,}` on exactly that reasoning — but four is not
where the reasoning ends. A cut that leaves one, two or three value characters is still under the
floor, so the sigil plus up to three characters of the secret still reach the log. The window is a
third of its old width, not closed.

This takes all three shapes to `{1,}`:

- `gh[pousr]_\w{1,}` and `github_pat_\w{1,}` (`GitHubApiError.java:93`)
- `bearer\s+[\w.~+/=-]{1,}` (`GitHubApiError.java:138`)

The sigil is the whole discriminator, which is the argument the javadoc above these patterns
already makes: `ghp_`, `github_pat_` and `Bearer ` do not occur in prose, so the length floor was
never carrying the discrimination. `{4,}` was a stopping point rather than a reasoned bound — no
test pinned it. One is where the argument genuinely ends: a cut that leaves *no* value characters
strands the sigil by itself, and a sigil is not secret.

The second half of the issue is the javadoc. `GitHubApiError.java:85-90` described the
severed-token case as closed while three characters were still going through. It now states the
residual it actually has, so the next reader starts from the true premise. The bearer shape's
paragraph at `:127-129` is updated to match.

### Over-masking

Every existing over-masking control stays green with no edit: `gh_2.40.0_linux_amd64.tar.gz`,
`ghost_writer`, `the ghp_ prefix and the ghs_ prefix`, the JWT-lookalike hostname and the
mid-run `eyJ` cases. Nothing in the suite regressed — full run below.

## Related Issues

Fixes #757

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

`GitHubApiErrorTest.ACredentialTheBoundCutJustShortOfItsLengthFloor` already pinned nine surviving
value characters for each of the three shapes. Its fixture helper is generalised to take the
surviving-character count, and three tests are added at **one** surviving character — the tightest
point of the floor, and the case the fix has to close.

### Red before the fix

The three new tests on the unfixed tree at `73cc33d` (verbatim, from the surefire XML report — the
console renders the trailing `…` as `?` because the surefire console stream is not UTF-8 here):

```
[ERROR] Tests run: 67, Failures: 3, Errors: 0, Skipped: 0
```

```
isStillMaskedForATokenPrefixCutToASingleCharacter
***,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,ghp_A… ==> expected: <false> but was: <true>

isStillMaskedForAFineGrainedPersonalAccessTokenCutToASingleCharacter
***,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,github_pat_A… ==> expected: <false> but was: <true>

isStillMaskedForABearerValueCutToASingleCharacter
***,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,Bearer S… ==> expected: <false> but was: <true>
```

That is the claimed mechanism exactly: the credential-shaped prefix ahead of the token compresses
to `***`, so the stranded sigil and its one surviving character clear the 512-character cap and
land in the `body=` field.

### Green after

`Tests run: 3328, Failures: 0, Errors: 0, Skipped: 0`.

One further test, `leavesTheBareSigilBehindWhenTheCutLandsBeforeTheFirstValueCharacter`, is a
**control** rather than proof — it is green before and after. It pins the residual the javadoc now
names: a cut landing before any value character leaves `ghp_…` in the line, which is the floor's
true boundary and carries nothing secret.

## Checklist

## Screenshots / Logs

Gates, in order:

- `./mvnw -B spotless:apply` — BUILD SUCCESS
- `./mvnw -B clean compile spotbugs:check spotless:check` — **BugInstance size is 0**, BUILD SUCCESS
- `./mvnw -B clean test` — **Tests run: 3328, Failures: 0, Errors: 0, Skipped: 0**

Coverage over `git diff -U0 73cc33d...HEAD` intersected with the jacoco report: **zero uncovered
lines and zero uncovered branches** in the changed main code.

## Additional Notes

Milestone v0.6.5, which is not tagged yet — this belongs in the release rather than on top of it.

#755 stacks on this branch, since it touches the same file; it re-targets `main` once this merges.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/LogSafe.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/LogSafe.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2026 Thiago Gonzaga
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package dev.thiagogonzaga.thrillhousebot;
+
+import java.util.regex.Pattern;
+
+/**
+ * The single place untrusted text is flattened before it is interpolated into a log line. Every
+ * value a log statement splices in that the bot did not choose itself — a GitHub error body, a
+ * model-supplied finding title or path — goes through here, so the "a field forges a record" class
+ * of defect is fixed in one place rather than re-litigated at each new call site (#731, #740,
+ * #742).
+ *
+ * <p>This is the log-destined counterpart to {@code MarkdownSafe}, and the two are deliberately
+ * separate. {@code MarkdownSafe.oneLine} feeds text the bot posts to GitHub, where the wider class
+ * below has a real rendering cost; a log record is a diagnostic identity rather than a rendering
+ * surface, and pays that cost gladly. Widening the markdown collapser instead would change what the
+ * bot posts.
+ */
+public final class LogSafe {
+
+  /**
+   * Collapses the whitespace of an untrusted string so one value stays on one log line.
+   *
+   * <p>Wider than {@code \s}, which java.util.regex reads as the ASCII six ({@code [
+   * \t\n\x0B\f\r]}) unless the pattern asks for Unicode character classes. CR and LF being
+   * collapsed closes the classic forged-record vector, but NEL (U+0085), LINE SEPARATOR (U+2028),
+   * PARAGRAPH SEPARATOR (U+2029), NUL and the ANSI escape all survived it (#731) — and a log
+   * viewer, a terminal, or a JSON/ECS shipper may treat any of them as a record boundary or as a
+   * screen-control sequence. A caller that reaches for this class has already decided its value is
+   * attacker-influenced text on its way to a log file and is already paying for a collapse pass on
+   * that basis; this is that pass covering what it claims to.
+   *
+   * <p>{@code \p{IsCc}} is the Unicode general category rather than POSIX {@code \p{Cntrl}}, so it
+   * reaches the C1 controls (U+0080–U+009F, NEL among them) as well as C0 and DEL.
+   *
+   * <p>{@code \p{IsZs}} covers the space separators {@code \s} leaves behind — NBSP (U+00A0), the
+   * EM/EN and figure spaces (U+2000–U+200A), the narrow NBSP (U+202F), the medium mathematical
+   * space (U+205F) and the ideographic space (U+3000). Without it the class contradicted the
+   * contract this method's javadoc states, and inconsistently: {@code String.strip()} reads {@code
+   * Character.isWhitespace}, which is true for U+2003 and U+3000 but false for U+00A0, U+2007 and
+   * U+202F, so the first pair were trimmed at the ends yet never collapsed inside, and the second
+   * three survived at every position. They forge no boundary, which is why this is the cheapest of
+   * the four classes to justify; they are here because two values differing only by one of them
+   * render identically in a log, which is the same harm as the invisible characters below.
+   *
+   * <p>{@code \p{IsCf}} is here for the same harm rather than for line integrity: bidi overrides
+   * and isolates (RLO, LRM, LRI) reorder what an operator reads, and the invisible joiners and
+   * spaces (ZWJ, ZWNJ, ZWSP, the BOM, the soft hyphen) let two different values render identically
+   * — both forge a record's meaning as surely as a forged boundary forges its extent. The accepted
+   * cost is that an echoed user string loses its grapheme clusters: an emoji ZWJ sequence or an
+   * Indic conjunct is split apart. A logged value is a diagnostic identity rather than a rendering
+   * surface, and which characters arrived is the question it exists to answer. Replacing with a
+   * space rather than deleting is part of the same bargain — deletion would let {@code
+   * admin<ZWSP>istrator} close up into a different real word, a space cannot.
+   */
+  private static final Pattern WHITESPACE =
+      Pattern.compile("[\\s\\p{IsZs}\\p{IsCc}\\p{IsCf}\\u2028\\u2029]+");
+
+  private LogSafe() {}
+
+  /**
+   * An untrusted string flattened to one log-safe line: every run of whitespace, control and format
+   * characters becomes a single space, and the ends are trimmed so a value that began or ended with
+   * such a run does not leave a stray space in the line. A {@code null} value flattens to the empty
+   * string, so a caller never has to guard for one.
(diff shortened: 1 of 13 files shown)


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/LogSafe.java:22 — LogSafe javadoc overclaims that every untrusted value in every log line goes through it

Raised on #760, which merged into this branch carrying it open. Verified and fixed here in `271efd6`.

The claim was false as written. All three lines interpolate `finding.file()` raw:

| line | statement |
|---|---|
| `ReviewPublisher.java:721` | `Line for %s:%d is outside the PR diff — filing the finding on the file instead` |
| `:729` | `Adjusted inline comment line for %s from %d to %d` |
| `:877` | `Inline comment rejected for %s:%d (suggestion=%s): %s` |

Widened the coverage rather than qualifying the claim. The sanitizing call is the one the warn lines beside them already make, and "debug is off by default" bounds who is exposed, not whether the vector exists — a debug-enabled install has precisely the surface the warn lines were fixed for. Qualifying the javadoc would have left a helper documenting its own gaps, which is the failure mode this PR exists to remove: `MarkdownSafe.oneLine` promised a collapse it did not perform, and #742 read as closed for a week because of it.

Grep confirms no `Log.*f` call in `ReviewPublisher` now interpolates `finding.file()` or `finding.title()` without going through the helper, so the class javadoc's "every value" is now true rather than aspirational.

This push also merges `main` (picking up #759) and carries the Zs widening from the earlier finding on #760.

Gates: `BugInstance size is 0`, spotless clean, full suite **3370 tests, 0 failures**.

---
_Generated by [Claude Code](https://claude.ai/code)_

devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewPublisher.java:771 — Log line reason slots interpolate rejectionReason raw despite LogSafe sweep

Investigated as asked. The finding pointed at the right slot; the mechanism turned out to be one field over, and there was a real defect there. Fixed in `7d46425`.

**On the question actually posed — can `rejectionReason` propagate raw response body text? No.** `rejectionReason` returns `GitHubApiError.diagnostics()`, whose body component is `body.logged()`, and `clean()` collapses that through `LogSafe` before redacting and capping it. The body reaches the reason slot already flattened.

**But `diagnostics()` also interpolates four response *headers*, and those were verbatim:**

```java
append(text, "retry-after", retryAfter);
append(text, "x-ratelimit-remaining", rateLimitRemaining);
append(text, "x-ratelimit-reset", rateLimitReset);
append(text, "x-ratelimit-resource", rateLimitResource);
```

They come from `response.getHeaderString(...)` and went into the line untouched — one field away from a body being collapsed for exactly this reason. `GitHubApiErrorHeaderCollapseTest`, red before the fix:

```
[ERROR] Tests run: 3, Failures: 3, Errors: 0, Skipped: 0
NEL survived into the diagnostics line:
  status=403 x-ratelimit-resource=core?forged WARN approved body={"message":"no"}
ESC survived into the diagnostics line:
  status=403 x-ratelimit-remaining=0?[2Kcleared body={"message":"no"}
LINE SEPARATOR survived into the diagnostics line:
  status=403 retry-after=30?forged body={"message":"no"}
```

The collapse now goes in `append`, the one place every header passes through, so a header added later is covered without anyone having to remember. Two further tests pin that the value is *kept* rather than blanked, and that an ordinary header is untouched.

**The `orElseGet(e::toString)` fallback** — used when the exception carried no response at all — was raw for the same reason and now takes the same treatment, as does the non-`WebApplicationException` return.

Worth recording that your note about the new test was correct and is why this needed a separate one: `ModelSuppliedTextInLogLinesTest` drives the path with a constant reason containing no control characters, so it could never have caught this.

Full suite **3376 tests, 0 failures**; `BugInstance size is 0`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #548, opened 2026-08-11
A low-confidence finding with no inline thread can never be cleared


devops-thiago:
Observed on PR #544, and it makes a clean PR look dirty.

Low-confidence findings below high risk do not post inline (`Finding#postsInline`); they are listed under "Things to double-check" in the summary. The deterministic backstop then holds them across rounds until they are resolved. But the only documented way for a maintainer to clear a finding is to fix it or reply on its review thread — and these findings have no thread. The bot's own message admits the gap in passing: *"fix them, or reply on their review thread (where one exists)"*.

So when such a finding IS fixed, the round that fixes it can still report it unresolved, because resolution depends on the model noticing the fix in a diff that may no longer touch those lines. On #544 a LOW was fixed in code and the next review still reported 2 previous findings unresolved, one of them that LOW.

Effects: a merge-ready PR displays an unresolved count that no action can clear, and APPROVE is held by the unresolved-previous gate.

Fix direction, to be decided with care: let a maintainer clear a thread-less finding from the PR conversation — the same marker + content recognition used for review-thread replies (#133/#142) — and make the unresolved message state that path instead of pointing only at threads. **Over-clearing is the dangerous direction**: a conversation comment must not clear findings it does not name, and must not clear findings from other rounds by reusing a marker. Prefer holding a finding over clearing it wrongly, and cover both directions with tests.

devops-thiago:
Already fixed on `release/v0.6.0` — closing as completed.

**Fixing PR:** #553
**Merge commit:** `eaa12b98` — confirmed an ancestor of the `release/v0.6.0` tip `fda4bc7` (`git merge-base --is-ancestor eaa12b98 fda4bc7`).

**Verified in the tree** (not taken from the PR description):
- `FollowUpAnalyzer` gained the conversation escape hatch: `CLEAR_DIRECTIVE`, `clearedInConversation`, `clearNamedInConversation`, `CONVERSATION_CLEARED_NOTE`, `namesLocator`/`LOCATOR_SHAPE` and `isMaintainerConversationComment` — so a low-confidence finding with no inline thread can be cleared from the PR conversation.
- Wired through `VerdictBuilder` (`ctx.conversationComments()`) and `ReviewContextLoader`; `GitHubCommentClient` now carries `author_association`; `MaintainerReplyService` distinguishes `CLEAR_DIRECTIVE_ACK` from `CLEAR_DIRECTIVE_NO_LOCATOR_ACK`.
[...]

Issue #512, opened 2026-08-10
Follow-up pass reports phantom "previous finding(s) remain unresolved" on a zero-finding PR


devops-thiago:
## Summary

Observed live on PR #511 (bot running v0.5.0): the first review pass **APPROVED with zero findings** — the summary comment says "No issues found in this PR", every risk count is 0, there are no inline comments and no "Things to double-check" block. The follow-up pass on the next revision then posted:

> No new issues in this revision, but 1 previous finding(s) remain unresolved — fix them, or reply on their review thread (where one exists) with why they are deferred.

There is no finding for that counter to refer to. The check-run conclusion was `neutral`, so nothing blocks — but the message tells a maintainer to fix something that does not exist, which costs exactly the trust the follow-up feature is meant to build.

## Likely mechanism (unverified — verify against the follow-up analyzer)

Candidates, in order of plausibility:

1. **Filtered-but-persisted findings haunt the tracker.** The persisted AI response (what `previousFindings` context is built from) may contain findings that the posting layer later filtered out (confidence gate, quote validation, dedup). A follow-up that counts "previous findings" from the persisted response, but "resolved/posted" from the PR surface, would see a phantom unresolved item that was never visible to anyone.
2. **Summary-only items counted as thread-requiring.** Related UX defect seen on #509/#510 the same day: "Things to double-check" items are explicitly *not* posted as inline threads, yet the unresolved-nag instructs "reply on their review thread (where one exists)" — for summary-only items no thread exists, so the instruction is unsatisfiable. If those items also feed the unresolved counter, a maintainer cannot ever clear it by the stated mechanism.

## What a fix needs to cover

- The unresolved counter must count only findings that were actually **posted** (inline thread or summary item), never persisted-but-filtered ones.
- A zero-findings prior pass must make the follow-up nag unreachable.
- For summary-only items, the nag must not instruct a thread reply that cannot exist — either give those items a resolvable surface or say how to acknowledge them.

## Provenance

Pre-existing (v0.5.0 production; the follow-up machinery predates this cycle) — **not** `rework`. Filed at v0.7.0: cosmetic-to-annoying, non-blocking (`neutral` check), does not affect v0.6.0 functionality.

devops-thiago:
Already fixed on `release/v0.6.0` — closing as completed.

**Fixing PR:** #470
[...]

PR #768, opened 2026-08-18
fix(review): converge on an anchor the maintainer has already litigated

### LINKED ISSUE #726: Re-reviews of an unchanged diff mint a new double-check finding every round instead of converging
Observed while landing #723: five consecutive review rounds over essentially unchanged code each produced a brand-new lower-confidence "Things to double-check" finding against the same ~30 lines (`GitHubApiError` wording patterns), and the rounds contradicted each other's fix direction:

1. "Content-creation wording rules only match the exact noun phrase" → gerund variant added.
2. "CONTENT_CREATION_BLOCK over-matches the generic secondary-rate-limit wording" → pattern scoped down (the opposite direction of round 1).
3. "Creation-block wording with an inserted word bypasses both throttle detection and the 30s floor" → declined as deliberate; loosening would re-open round 2.
4. "30s floor also covers 'abuse detection' wording — verify it is the same long block" → refuted from GitHub's pre-2021 wording history.
5. "Block-recognition regex is a fixed word-order whitelist; a reworded block message silently drops the 30s floor" (MEDIUM) → premise ("no log change to reveal it") factually contradicted by the same PR's logging half.

Each round's finding was novel by title/anchor, so previous-findings tracking never converged: every `/review` was a fresh roll of speculative "input not in the diff" hypotheticals against the same lines, and the operator had to disposition each one manually via `@thrillhousebot resolved`. This is distinct from #512 (phantom unresolved): the findings are new inventions per round, not stale carryovers.

Direction ideas (not prescriptive):
- Feed the model prior rounds' *declined/justified* double-check findings (with their justifications) as context for re-reviews of an unchanged diff, with an instruction that a new finding on the same anchor must engage the recorded justification rather than restate a variant of it.
- Suppress or down-rank new lower-confidence findings on lines whose prior findings were maintainer-declined this PR, or require the model to mark them as disputing the recorded disposition.
- Consider a per-anchor cap: after N maintainer-declined findings on the same `path:line`, further lower-confidence findings on that anchor are listed as "previously litigated" rather than counted open.

Evidence chain: PR #723 review timeline (reviews 2026-08-15 17:31 → 22:58 and the four `@thrillhousebot resolved` dispositions).

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [ ] ✨ Feature
- [ ] 📝 Documentation
- [ ] 🔧 Refactor
- [ ] 🚀 Performance
- [ ] ✅ Test
- [ ] 🔒 Security
- [ ] 📦 Dependency update
- [ ] 🏗️ CI/CD

> **Stacked on #766** (`fix/736-bound-pr-context`). Review the top commit; the base retargets to `main` when the parent merges.

## Description

Five consecutive rounds over essentially unchanged code each produced a brand-new lower-confidence "Things to double-check" item against the same ~30 lines, and the rounds contradicted each other — round 2 argued the opposite of round 1, round 5's premise was refuted by the same PR's other half. Because every item was novel by title and premise, previous-findings tracking had nothing to converge on: each `/review` was a fresh roll of speculative hypotheticals against the same lines, and the maintainer had to disposition a new one every round.

`FollowUpAnalyzer#dropRepliedDuplicates` already stops a prior finding from being *re-raised*, but it matches on the **defect** (title similarity or content overlap), and a new title with a new premise on the same lines is not the same defect, so it passes straight through. The new pass matches on the **anchor** instead — the one thing the rounds actually have in common.

**The rule.** Once `LITIGATED_ANCHOR_LIMIT` (2) *distinct* findings on one `path:line` have been dispositioned by a maintainer on this pull request, a further finding on that anchor is withheld, with an `INFO` naming the finding, the anchor and the count.

**Which of the issue's three directions, and why.** The third — a deterministic per-anchor cap. The first (feed the declines and their justifications back as prompt context with an instruction to engage them) is more of what already exists and did not converge: the block already carries the prior findings, their threads and an "answered in earlier rounds — do NOT raise these again" instruction, and the model still minted a variant every round; adding prose asks the same model to be persuaded by the same evidence, and it costs shared overhead on every batch call. The second (down-rank) does not help either, because the class in question is *already* down-ranked — these items are exactly the ones confidence-routing keeps off the diff, and they still hold approval and still need a manual disposition, which is the reported cost. Only a deterministic rule converges, and the anchor is the deterministic thing.

**Why it cannot suppress real review**, by construction:

- **Only findings that never post inline are eligible.** `Finding#postsInline` routes low-confidence findings below high risk to the summary, and that is the whole reported class. A high-confidence finding, and any critical/high-risk finding, is untouched however often its anchor has been argued over — the constraint the issue sets.
- **Distinct findings only.** One defect carried across rounds and cleared in each is one argument, not two (`isSameFinding` clusters them), so the limit cannot be reached by repetition of a single item.
- **A disposition means a maintainer acted** — a reply on the finding's own thread, or an `@thrillhousebot resolved` comment naming it on the PR conversation, which is the only hatch a threadless finding has (#548). The bot's own rounds never raise the count.
- **The limit is two**, so the first decline settles one hypothesis and a second look at the same lines is still ordinary review. It is the third and later ones that stop being review.

**The trade it accepts:** if the code under a twice-litigated anchor genuinely changes, a new lower-confidence observation there is withheld too. That is the cheap direction — the same defect surfacing at high confidence or at critical/high risk is exempt, and the withheld class is by construction the one the review already declines to put on the diff. Withholding is disclosed to the operator in the log, the way the sibling `dropRepliedDuplicates` drop is; no new review surface is introduced, since giving the noise a section would be the opposite of converging.

Applied at both `FindingVerificationService`-adjacent seams in `FindingPipeline`, immediately after `dropRepliedDuplicates`, so the two passes read as the pair they are.

## Related Issues

Fixes #726

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

**Red/green proof.** `FindingPipelineTest#aThirdSpeculativeItemOnATwiceDispositionedAnchorIsWithheld` runs the real pipeline over a context with two prior rounds whose findings the maintainer cleared with `@thrillhousebot resolved`, and a current round raising a third, novel, lower-confidence item on the same lines plus an unrelated finding elsewhere. Run against this branch's test with the main-code change reverted:

```
[ERROR] Tests run: 2, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 1.260 s <<< FAILURE! -- in dev.thiagogonzaga.thrillhousebot.review.FindingPipelineTest
[ERROR] dev.thiagogonzaga.thrillhousebot.review.FindingPipelineTest.aThirdSpeculativeItemOnATwiceDispositionedAnchorIsWithheld -- Time elapsed: 1.186 s <<< FAILURE!
org.opentest4j.AssertionFailedError: a third lower-confidence item on the twice-dispositioned anchor must not be raised ==> expected: <[Unrelated defect elsewhere]> but was: <[Reworded message drops the 30s floor, Unrelated defect elsewhere]>
	at org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1210)
	at dev.thiagogonzaga.thrillhousebot.review.FindingPipelineTest.aThirdSpeculativeItemOnATwiceDispositionedAnchorIsWithheld(FindingPipelineTest.java:380)
```

With the fix it passes, and the unrelated finding is kept.

`FindingPipelineTest#aHighConfidenceFindingOnATwiceDispositionedAnchorStillPosts` is a **control**, not proof — it passes before and after, and pins the boundary the rule must never cross.

`FollowUpAnalyzerTest` adds eight focused cases on the new pass: the conversation-clear signal, the thread-reply signal, an inline-posting finding on the same anchor, one disposition only, one defect argued twice counting once, other files and distant lines, no disposition at all, and the empty inputs. The previous-findings accounting and the recounted summary are asserted alongside.

Gates on this branch: `spotless:apply` clean, `clean compile spotbugs:check spotless:check` → **BugInstance size is 0**, `clean test` → **3394 tests, 0 failures, 0 errors**. Jacoco ∩ `git diff -U0 c4550f8...HEAD` shows zero uncovered lines and zero uncovered branches in the changed main code (checked against the parent commit and against `c4550f8`).

## Checklist

## Screenshots / Logs

```
INFO  [dev.thiagogonzaga.thrillhousebot.review.FollowUpAnalyzer] Withholding lower-confidence finding 'A reworded block message drops the 30s floor' (src/A.java:12) — 2 finding(s) on that anchor were already dispositioned by a maintainer on this pull request
```

## Additional Notes

This is distinct from #512 (phantom unresolved): those are stale carryovers, these are new inventions per round. The pass touches neither previous-findings statuses nor the verdict — it only decides whether this round raises the item.

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingPipeline.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingPipeline.java
@@ -450,6 +450,15 @@ public class FindingPipeline {
     refined =
         followUpAnalyzer.dropRepliedDuplicates(
             refined, ctx.priorAiResponseJsons(), ctx.inlineComments(), botIdentity);
+    // #726: a novel lower-confidence hypothesis on an anchor the maintainer has already
+    // dispositioned twice is a re-roll of the dice, not a finding to disposition again.
+    refined =
+        FollowUpAnalyzer.withoutPreviouslyLitigated(
+            refined,
+            ctx.priorAiResponses(),
+            ctx.inlineComments(),
+            ctx.conversationComments(),
+            botIdentity);
     refined = populateMissingAnchors(refined, lineResolver);
 
     if (tokenLedger.ceilingReached(ledgerSessionId(session))) {
@@ -1351,6 +1360,15 @@ public class FindingPipeline {
     aiResponse =
         followUpAnalyzer.dropRepliedDuplicates(
             aiResponse, ctx.priorAiResponseJsons(), ctx.inlineComments(), botIdentity);
+    // #726: a novel lower-confidence hypothesis on an anchor the maintainer has already
+    // dispositioned twice is a re-roll of the dice, not a finding to disposition again.
+    aiResponse =
+        FollowUpAnalyzer.withoutPreviouslyLitigated(
+            aiResponse,
+            ctx.priorAiResponses(),
+            ctx.inlineComments(),
+            ctx.conversationComments(),
+            botIdentity);
     aiResponse = populateMissingAnchors(aiResponse, lineResolver);
     persistAiResponse(session, aiResponse);
     return aiResponse;
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FollowUpAnalyzer.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FollowUpAnalyzer.java
@@ -739,7 +739,8 @@ public class FollowUpAnalyzer {
       return response;
     }
     var dispositioned =
-        dispositionedFindings(priorRounds, inlineComments, conversationComments, botIdentity);
+        distinct(
+            dispositionedFindings(priorRounds, inlineComments, conversationComments, botIdentity));
     if (dispositioned.isEmpty()) {
       return response;
     }
@@ -797,15 +798,34 @@ public class FollowUpAnalyzer {
    * rounds carried it, so it is counted once.
    */
   private static int litigating(
-      ReviewResponse.Finding finding, List<ReviewResponse.Finding> dispositioned) {
-    var counted = new ArrayList<ReviewResponse.Finding>();
-    for (var prior : dispositioned) {
-      if (sameAnchor(finding, prior)
-          && counted.stream().noneMatch(seen -> isSameFinding(prior, seen))) {
-        counted.add(prior);
+      ReviewResponse.Finding finding, List<ReviewResponse.Finding> distinctDispositioned) {
+    var count = 0;
+    for (var prior : distinctDispositioned) {
+      if (sameAnchor(finding, prior)) {
+        count++;
       }
     }
-    return counted.size();
+    return count;
+  }
+
+  /**
+   * {@code dispositioned} with findings that are the same defect collapsed to one, so a defect the
+   * maintainer dispositioned once and the model then re-raised across rounds counts once toward the
+   * limit rather than once per round.
+   *
+   * <p>Collapsed here, once per call, rather than inside {@link #litigating}: that runs per finding
+   * in the response, and folding the same list again for each of them repeated the whole comparison
+   * for no new answer. {@link #isSameFinding} already requires the same file and a line within
+   * tolerance, so collapsing across the list and collapsing within one anchor give the same groups.
+   */
+  private static List<ReviewResponse.Finding> distinct(List<ReviewResponse.Finding> dispositioned) {
+    var distinct = new ArrayList<ReviewResponse.Finding>(dispositioned.size());
+    for (var prior : dispositioned) {
+      if (distinct.stream().noneMatch(seen -> isSameFinding(prior, seen))) {
+        distinct.add(prior);
+      }
+    }
+    return distinct;
   }
 
   /** Whether two findings sit at the same location, within the drift a revision may introduce. */
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FollowUpAnalyzer.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/FollowUpAnalyzer.java
@@ -682,6 +682,138 @@ public class FollowUpAnalyzer {
     return null;
   }
 
+  /**
+   * How many distinct findings on one anchor a maintainer must have dispositioned before a further
+   * lower-confidence finding on that same anchor stops being raised (#726).
+   *
+   * <p>Two, not one: the first decline settles one hypothesis, and a second look at the same lines
+   * is ordinary review. It is the third and later ones that stop being review and start being a
+   * fresh roll of the dice — the observed run was five rounds over essentially unchanged code, each
+   * minting a brand-new lower-confidence item on the same ~30 lines, each contradicting the last,
+   * and each costing the maintainer a manual disposition.
+   */
+  static final int LITIGATED_ANCHOR_LIMIT = 2;
+
+  /**
+   * Deterministic convergence for an anchor the maintainer has already litigated: withholds a new
+   * <em>lower-confidence</em> finding whose location has already carried {@link
+   * #LITIGATED_ANCHOR_LIMIT} distinct maintainer-dispositioned findings on this pull request
+   * (#726).
+   *
+   * <p>{@link #dropRepliedDuplicates} already stops a prior finding from being re-raised, but it
+   * matches on the defect: a new title with a new premise on the same lines is not the same
+   * finding, so it passes. That is exactly what re-review of an unchanged diff produces — each
+   * round invents a novel speculative hypothesis against the same lines, previous-findings tracking
+   * never converges because nothing is repeated, and the maintainer disposes of a new item every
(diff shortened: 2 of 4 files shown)


devops-thiago:
Three findings, one valid. Addressed in `1ecc93c`.

### 🔵 LOW — `litigating` dedupes with a linear scan per element — valid, fixed

The premise needs one correction: `sameAnchor` filters *before* the scan, so `counted` only ever accumulates same-anchor findings and the cost is O(d·k), not O(d²) in general. But k → d in exactly the case this feature exists for — many findings litigated on one anchor — so the distinction does not save it.

Fixed the way suggested: the collapse is hoisted to one call in `withoutPreviouslyLitigated`, and `litigating` now counts anchor matches in a single pass. `isSameFinding` already requires the same file and a line within tolerance, so collapsing across the whole list and collapsing within one anchor produce the same groups — the counts are unchanged, which the existing `FollowUpAnalyzerTest` cases and the pipeline tests pin (219 and 73 green, unmodified).

### 🟡 MEDIUM — `sameAnchor` drops the null-file guard `isSameFinding` has — reviewer mistake

No NPE is reachable. `sameAnchor` delegates to `FilePaths.same`, which opens with:

```java
static boolean same(String a, String b) {
  if (a == null || b == null) {
    return false;
  }
```

`isSameFinding`'s explicit `finding.file() == null` test short-circuits before the same call, so the two are behaviourally identical on a null file — both yield `false`. The guard in `isSameFinding` is redundant with its own callee, not a protection `sameAnchor` is missing.

### 🟡 MEDIUM — `PriorAiResponses` wiring unverified, may never fire in production — reviewer mistake

`ctx.priorAiResponses()` is not new plumbing. It is populated in `ReviewContextLoader` from `parsePreviousResponses(priorAiResponseJsons)` — the persisted prior rounds — and is already consumed in production by `VerdictBuilder.settledPreviousIds` and the approve backstop. This change reads the same field at the same seam.

The tests drive it at the pipeline level rather than by calling the method directly, which is what the red proof shows: `FindingPipelineTest.aThirdSpeculativeItemOnATwiceDispositionedAnchorIsWithheld` failed with `[Reworded message drops the 30s floor, Unrelated defect elsewhere]` before the fix and passes after, having gone through `FindingPipeline` and the real context.

### Gates

`BugInstance size is 0`, full suite **3394 tests, 0 failures**, jacoco ∩ diff **55 executable changed lines, zero uncovered lines and zero uncovered branches**.

devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/FollowUpAnalyzer.java:813 — sameAnchor drops the null-file guard that sibling isSameFinding has

Verified as the finding asks: `FilePaths.same` is null-tolerant, so this is inert.

```java
static boolean same(String a, String b) {
  if (a == null || b == null) {
    return false;
  }
  return a.equals(b) || suffixAtBoundary(a, b) || suffixAtBoundary(b, a);
}
```

A null `file` on either side yields `false` and `sameAnchor` returns `false`. No dereference happens, so `withoutPreviouslyLitigated` cannot throw on that input and the fail-soft requirement holds.

The reasoning from the sibling guard is fair but reads it backwards: `isSameFinding`'s explicit `finding.file() == null` test short-circuits before a call that already handles null, so it is redundant with its own callee rather than a protection `sameAnchor` is missing. Both methods return `false` for a null file; they differ only in where the check sits.

The observation that the pass runs before `populateMissingAnchors` is correct and is why a null `file` genuinely reaches here — it just reaches a null-safe comparison.

devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/FindingPipeline.java:458 — PriorAiResponses wiring unverified — fix may never fire in production reviews

Verified in `ReviewContextLoader`, which the finding correctly notes was not in the provided material. Both fields are derived from the same source that feeds `priorAiResponseJsons()`, so they cannot be populated apart from it.

**Parsed prior rounds** — the same list, parsed once:

```java
List<String> priorAiResponseJsons = ...;                       // from fetchPriorReviews
var hasContext = !priorAiResponseJsons.isEmpty();
List<ReviewResponse> priorAiResponses =
    followUpAnalyzer.parsePreviousResponses(priorAiResponseJsons);
```

**Conversation comments** — gated on that same emptiness check:

```java
List<GitHubCommentClient.IssueComment> conversationComments =
    hasContext
        ? fetchConversationComments(auth, req.owner(), req.repo(), req.prNumber())
        : List.of();
```

So the two fields the pass consumes are populated exactly when `priorAiResponseJsons` is non-empty — which is the fix's trigger. The empty case is a first review, where there are no prior rounds to have litigated anything and the pass correctly returns the response unchanged.

The scenario the finding describes — the loader filling `priorAiResponseJsons()` while leaving `priorAiResponses()` or `conversationComments()` empty — is not reachable, because `hasContext` is computed from the first and gates the other two.

Worth adding that `ctx.priorAiResponses()` is not new plumbing this PR introduced: `VerdictBuilder.settledPreviousIds` already reads it in production, and `ReviewContextLoader` derives `previousFindingsSource` from it via `effectivePreviousFindings`.

The point about the test constructing its context by hand is right, and is the reason this needed checking against the loader rather than against the test.

PR #783, opened 2026-08-24
fix: disclose patchless files under their own reason, not the review budget (#628)

### LINKED ISSUE #628: Patchless files are disclosed as having exceeded the review budget

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix

## Summary
- A reviewable file GitHub reports with real additions/deletions but no patch text (binary, or a text diff too large to display) was folded into the plan's budget-omitted list, so the posted review told the maintainer it "exceeded the review budget" — pointing at a budget knob that would change nothing.
- Following the direction recorded in the issue, the omission *reason* is now carried rather than inferred: `DiffBudgetPlanner.BudgetPlan` gains a `patchlessFiles` component and `ReviewResult.TruncationDetail` gains `patchlessFileNames`, each with a convenience constructor preserving the old canonical signature so the call sites were updated in one deliberate pass without churn.
- Every disclosure surface now names the honest cause — "GitHub provided no diff content … binary files, or text diffs too large to display": the review banner / coverage-gap clause, the check-run brief, the on-demand command disclosure (`/describe`, `/changelog`, …), the summary model's changed-files overview, and the withheld-material notice in the batch prompt.
- Behavior is otherwise unchanged: patchless files still hold APPROVE, still drop their walkthrough rows, still trigger the degenerate skip-the-review-call path when they are the only reviewable files, and are never packed into a batch.

## Test plan
- [x] Planner tests updated: patchless files land in `patchlessFiles`, not `omittedFiles`, and still make the plan truncated (`DiffBudgetPlannerTest`)
- [x] Verdict tests: patchless disclosed under its own clause, budget wording absent, coexistence with true budget omissions, check-run brief label (`VerdictBuilderTest`)
- [x] Rendering tests: coverage-gap clause/brief, disclosure for patchless-only details, null tolerance (`ReviewResultTest`)
- [x] Pipeline tests: withheld-material notice and summary overview name the real reason; patchless-only plan skips the review call (`FindingPipelineTest`)
- [x] `./mvnw clean compile spotbugs:check spotless:check` and `./mvnw clean test` (3422 tests) pass locally; JaCoCo shows zero uncovered changed lines

## Related Issues

Fixes #628

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [x] Manual testing

Local gates: `spotless:check`, `spotbugs:check`, full test suite pass; CI green except review follow-ups in progress.

## Checklist

## Screenshots / Logs

N/A

### DIFF
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/AbstractPrSuggestionGenerator.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/AbstractPrSuggestionGenerator.java
@@ -256,22 +256,26 @@ public abstract class AbstractPrSuggestionGenerator {
 
   /**
    * The partial-coverage disclosure for a run, sourced from the token-budget plan: files that did
-   * not fit any batch and files clipped to fit one. Both are named, so a large PR says which parts
-   * of the change set the command did not cover rather than only how many — and coverage is now
-   * bounded by {@code max-ai-calls}, never by the diff render's line cap.
+   * not fit any batch, files clipped to fit one, and files GitHub returned no patch text for
+   * (#628). All are named, so a large PR says which parts of the change set the command did not
+   * cover rather than only how many — and coverage is now bounded by {@code max-ai-calls}, never by
+   * the diff render's line cap.
    */
   protected static String disclosure(DiffBudgetPlanner.BudgetPlan plan) {
     if (!plan.truncated()) {
       return "";
     }
     return ReviewResult.truncationDisclosure(
-        plan.omittedFiles().size() + plan.clippedFiles().size(),
+        plan.omittedFiles().size() + plan.clippedFiles().size() + plan.patchlessFiles().size(),
         new ReviewResult.TruncationDetail(
             plan.omittedFiles(),
             plan.clippedFiles(),
+            plan.patchlessFiles(),
             List.of(),
             List.of(),
-            SummaryDegradation.NONE));
+            List.of(),
+            SummaryDegradation.NONE,
+            VerificationCoverage.EMPTY));
   }
 
   /**
--- a/src/main/java/dev/thiagogonzaga/thrillhousebot/review/DiffBudgetPlanner.java
+++ b/src/main/java/dev/thiagogonzaga/thrillhousebot/review/DiffBudgetPlanner.java
@@ -108,11 +108,18 @@ public class DiffBudgetPlanner {
    * but every batch that succeeded keeps its findings. Written only through {@link
    * #recordUncoveredFiles(List)}; its accessor returns a defensive copy so the mutable backing
    * never escapes. The {@code effective*} accessors fold it into the planned omissions.
+   *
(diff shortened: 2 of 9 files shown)


devops-thiago:
Re the two "Things to double-check" items from the bot review:

1. **TruncationDetail.isEmpty() (MEDIUM)** — not a gap: `isEmpty()` does not enumerate the file lists, it delegates to `hasFileGaps()` (`return !hasFileGaps() && summaryDegradation == NONE && !verification.disclosed();`), which this PR extends with `patchlessFileNames`. A patchless-only detail therefore already returns `isEmpty() == false`. Added a regression assertion (`truncationDisclosureRendersForAPatchlessOnlyDetail`) pinning that behavior.

2. **Zero-diff-budget early exit in renderAndSize (LOW)** — ruled out: the `diffBudgetTokens <= 0` branch is the explicit `max-input-tokens=0` legacy no-cap path. It adds every file to the single uncapped batch (`rendered.sized().add(...)`) and never writes `unclippable`/omitted, so no budget-blaming disclosure can be produced there; the budgeted overload floors the diff budget at 1, so a budgeted plan never reaches that branch. Pre-existing behavior, unchanged by this PR.

The one new Sonar issue (S3024, string concat inside `append`) is fixed by extracting the note into a constant.

devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/DiffBudgetPlanner.java:758 — Patchless classification may be bypassed on the zero-diff-budget early-exit path

PR #790, opened 2026-08-24
fix(review): read ignore globs as the gitignore syntax they are documented as

### LINKED ISSUE #481: fix(review): make per-repo ignore globs behave as their documented gitignore-style syntax, and harden the YAML parser
### Problem Statement

An audit of the v0.6.0 repo-settings subsystem (`#51` per-repo ignore globs, `#33` path-scoped
instructions) found that the matcher does not behave the way its own documentation promises, plus
three narrower parser gaps. All were reproduced against `release/v0.6.0` and independently
validated. None is a security or verdict-correctness defect, so they are deferred here rather than
held against the v0.6.0 release — but the first one fails **open** (paths a repository asked to
exclude are sent to the model anyway), so it is worth fixing before v0.7.0.

### 1. "gitignore-style" is documented but not implemented — common forms silently match nothing

`README.md` (the per-repo section, new this milestone) and `RepoSettings`'s javadoc tell maintainers
per-repo globs use "the same gitignore-style syntax as the global key". The matcher is raw Java NIO
`FileSystems.getDefault().getPathMatcher("glob:" + pattern)` plus a suffix matcher built **only** when
the pattern starts with `**/`. Under that matcher, gitignore idioms silently match nothing:

| pattern a maintainer writes | intended | actual |
|---|---|---|
| `build/` (trailing slash = directory) | exclude the tree | matches nothing |
| bare `vendor` / `payments` | exclude the dir | matches nothing (matches only a file literally named that) |
| `*.lock` | every lockfile | matches only a **root-level** `*.lock` |
| `generated/**` | the dir anywhere | matches nothing (needs the `**/` prefix) |

`compileGlobMatchers` warns only on patterns that fail to **compile** (`InvalidPathException` /
`PatternSyntaxException`); `build/` compiles fine, so there is no signal at any log level, and a
path-scoped glob that matches nothing is dropped with no log at all. The shipped global default list
(`**/build/**`, `**/vendor/**`, `**/*.lock`) covers the common cases, so the harm is limited to
**novel** directory names a repository excludes (`generated/`, `payments/`, `fixtures/`, `testdata/`).

**Decision (maintainer): make the code match the docs**, not walk back the promise. Normalize a
per-repo pattern before compiling — strip a trailing `/` and append `/**`; when a pattern contains no
`/`, also compile it as `**/<pattern>` (the same helper the `**/` prefix already uses). This extends
the existing `**/`-suffix fallback, which already papers over one of the four idioms. Surface unmatched
declarations too: a one-line note in the review summary ("N declared ignore globs matched no file in
this PR") turns a silent miss into a fixable one.

### 2. Two more silent-drop paths in the same parser

- A **scalar** `ignored-files` value is comma-split (`RepoSettingsParser`), which destroys a brace
  glob: `ignored-files: "**/*.{js,ts}"` becomes `[**/*.{js, ts}]`, and both fragments then fail to
  compile and are dropped. The deployment-side env-var list is comma-separated too and has the same
  hazard.
- A single **duplicate key anywhere in the file** (`setAllowDuplicateKeys(false)`) throws and
  discards **every** setting, warn-logged only — a far wider blast radius than the parser's otherwise
  careful per-entry "one bad entry costs only itself" design.

### 3. YAML anchors/aliases are not resolved

`jackson-dataformat-yaml` never runs snakeyaml's `Composer`, so `*common` reaches the parser as the
literal string `"common"`. An aliased ignore list becomes a glob matching a file named `common`; an
aliased scope's `instructions` becomes a short literal string that passes every parser check and is
rendered into the review prompt as the repository's rule for those files. Merge keys (`<<: *base`)
fail closed (the entry loses its `instructions` and is dropped, with a warning). Fix: either reject a
document containing `&`/`*` anchors with a clear warning, or pre-resolve with snakeyaml's Composer
path and hand the object graph to Jackson's `convertValue`.

### 4. Two snakeyaml loader guards are inert, and their javadoc claims protection that does not exist

`setNestingDepthLimit(20)` and `setMaxAliasesForCollections(50)` are enforced in the Composer, which
Jackson never runs — they do nothing. The real nesting bound is jackson-core's `StreamReadConstraints`
default of 1000, which **is** caught and degrades to `EMPTY`, so this is a false invariant, not an
exploitable hole. `MAX_NESTING_DEPTH` / `MAX_ALIASES` are dead constants whose comments would let a
future reader believe the parser is hardened where it is not. Delete them (or set the depth via
`StreamReadConstraints`), and fix the comments. **Coupling note:** if `#3` is fixed by actually
resolving aliases, `setMaxAliasesForCollections` becomes load-bearing — fix them together.

### 5. Transient GitHub failure is negative-cached like a 404 (narrower, related)

`RepoSettingsResolver.fetchAndParse` catches `WebApplicationException | ProcessingException` and
returns `null`; `WebApplicationException` is the parent of `ServerErrorException` (500/502/503) and
`ClientErrorException` (403 secondary rate limit), so a transient failure is indistinguishable from a
real 404 and gets `RepoSettings.EMPTY` cached for `NEGATIVE_CACHE_TTL_MS` (60s). For that window every
review of the repo runs on the deployment ignore list only. Fix: catch `NotFoundException` separately
and negative-cache only that; for any other failure return without writing the cache, and log at
`warn` rather than the current "config file not found" `debug`.

### Acceptance criteria

- [ ] A per-repo pattern of `build/`, bare `vendor`, `*.lock`, or `generated/**` excludes what a
      gitignore-literate maintainer expects — with tests written the way a maintainer writes them
      (not `**/`-prefixed)
- [ ] A brace glob and a scalar comma-list are handled without silently dropping fragments
- [ ] A duplicate key degrades to a per-entry drop, not a whole-file discard
- [ ] A YAML alias either resolves correctly or is rejected with a clear warning
- [ ] The inert nesting/alias guards are removed or made real, and their comments are honest
- [ ] A transient 5xx/403 does not get negative-cached as "no config"
- [ ] Unmatched ignore globs are disclosed to the maintainer rather than silently ignored

### Environment

All present on `release/v0.6.0`. The gitignore-syntax wording and the `RepoSettings.path` javadoc are
new this milestone; the parser guards and the negative-cache behavior predate it. Found in the v0.6.0
deep audit; deferred from the release under the "regressions + live security ship now, latent hardening
to v0.7.0" scope decision.

### Code of Conduct

- [x] I have searched for existing issues and this is not a duplicate.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [x] 📝 Documentation

## Description

The per-repo and deployment-wide ignore keys are documented as "gitignore-style", but the
matcher was raw Java NIO globs plus a `**/`-prefix fallback. Under it the common idioms
silently matched nothing, verified against `main` before the fix:

| pattern a maintainer writes | before | now |
|---|---|---|
| `build/` | matches nothing | the `build` tree at any depth |
| bare `vendor` | only a file literally named `vendor` | a file or directory `vendor`, and its tree, at any depth |
| `*.lock` | only a root-level lockfile | every lockfile at any depth |
| `/Makefile` | matches nothing (absolute path) | the root `Makefile` only |
| `!path` | a matcher for a file named `!path` | dropped with a warning |

Each of those fails **open** — the file the repository asked to exclude is sent to the
model anyway — and fails silently, because the compile-time warning only fires on a
pattern that fails to *compile*. A pattern that matches nothing is indistinguishable from
one that had nothing to match.

Declared patterns are now normalized to gitignore's own reading before compiling
(`gitignoreForms`, one pure function): a leading `/` anchors and is dropped, a `/` anywhere
else anchors to the root, anything else matches at any depth, a trailing `/` means that
directory's tree, and a bare name means a file *or* a directory. The deployment key, a
repository's own list, and the path-scoped rule globs all compile through the one matcher,
so the "same syntax" promise is structural rather than a claim.

The issue's fourth row, `generated/**`, was the one claim that did not reproduce: it already
matches the root-level tree today. gitignore anchors a pattern carrying a slash, and that is
also the only reading under which the documented `docs/generated/**` example keeps meaning
what it says, so it stays anchored and `**/generated/**` remains how you say "at any depth".
The issue's own prescribed fix ("when a pattern contains no `/`, also compile it as
`**/<pattern>`") is exactly that rule.

**Disclosure.** The other half of a dead glob is that nobody finds out. The summary's
review-scope note now names the repository's globs that matched no file in this PR, sharing
the blockquote with the pure-rename rollup that already answers the same question.

**Parser gaps (issue items 2–4), fixed together.**

- A brace glob survived neither the scalar comma-split (`**/*.{js,ts}` became `**/*.{js` and
  `ts}` — the first uncompilable, the second a matcher for a file named `ts}`) nor the review.
  The split now ignores commas inside a brace group.
- A single duplicate key threw out of the whole parse and discarded every setting in the
  file. It is warned about, then re-read with YAML's usual last-one-wins.
- Aliases were never resolved: jackson-dataformat-yaml drives snakeyaml's event parser and
  never runs its composer, so `*common` arrived as the literal string `common` — an aliased
  ignore list became a glob for a file called `common`, and an aliased scope path became a
  short literal that passed every check and reached the review prompt as the repository's
  rule for those files. The document is loaded through snakeyaml's own loader now, which also
  makes `setNestingDepthLimit` and `setMaxAliasesForCollections` real; they were inert, with
  javadoc claiming a protection that did not exist, which is why the issue asks for them in
  the same change.

**Item 5.** A transient GitHub failure is no longer negative-cached like a 404.
`WebApplicationException` is the parent of `ServerErrorException` (5xx) and of the 403
secondary rate limit as well as of `NotFoundException`, so a blip pinned "this repository has
no config" for a minute of reviews. Only a real 404 is cached; anything else runs on the
global list for that review and is re-asked on the next.

Nothing here can widen review scope: the effective ignore set is still global ∪ per-repo, and
negation is refused precisely because it would subtract from a union.

## Related Issues

Closes #481

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

Every behavioural change was written test-first and run against the unfixed code. The six new
gitignore-idiom tests failed with `expected: <true> but was: <false>` (and the negation one
`expected: <false> but was: <true>`); the parser tests failed with `expected: <[generated/**,
**/*.snap]> but was: <[shared]>`, `expected: <payments/**> but was: <money>`, `expected:
<[kept/**]> but was: <[]>` for the duplicate key, `expected: <[**/*.{js,ts}]> but was:
<[**/*.{js, ts}]>` for the brace glob, and the cache test with "a failure we cannot read is
not an answer to cache". The scope-glob and disclosure tests were proven red against a
temporarily reverted matcher/wiring.

The shipped default ignore list is pinned by a test that reads the `@WithDefault` value
itself, asserting `**/pom.xml` reaches every pom at every depth and matches no neighbour that
merely contains the name, plus a sample across every other class of default entry — a change
in matcher semantics cannot quietly alter what the defaults exclude.

Gates: `spotless:apply`, `clean compile spotbugs:check spotless:check` (BugInstance size is
0), `clean test` — 3498 tests, 0 failures, 0 errors. Patch coverage over `git diff -U0
origin/main`: 140 instrumented changed lines, 0 uncovered lines, 0 uncovered branches.

**Review follow-up: what a readable-but-unusable config does to the cache.** A review read
the `Fetched` javadoc's "cacheable" as contradicting `resolve()`, on the reading that the
found branch returns before any `cache.put`. It does not — the write is inside the loop, and
`RepoSettingsParser.parse` never returns `null`, so a file that was read but parsed to nothing
takes the found branch and *is* cached. The javadoc was right; it just did not say where the
write happens, so it now does.

The behaviour is kept as it stands, and the judgement is worth stating rather than assuming.
A malformed config is cached for the full `CACHE_TTL_MS`, not the shorter negative TTL,
because the repository answered — it just said nothing usable. That does mean a maintainer
who fixes broken YAML waits out the TTL, but that is precisely the same wait as a maintainer
who edits YAML that already parsed, so the broken case is not treated worse than the working
one; and the alternative, leaving it uncached, spends a GitHub read on every review, forever,
for exactly the repositories that get no benefit from it. Two tests now pin both halves (it is
cached at all, and on the long TTL) — they fail with `expected: <1> but was: <0>` against a
copy of the code with the in-loop `cache.put` removed, so the claim is tested rather than
asserted.

**SonarCloud.** Two issues on this branch, both fixed rather than suppressed.
`Fetched.UNREADABLE` differed only by case from the record's own `unreadable` component
(`java:S1845`), which is a genuine trap in a type whose whole job is telling three outcomes
apart; the constant is renamed `READ_FAILED`, keeping the accessor, since `unreadable()` is
how `CiStatusEvaluator.CiEvaluation` already spells this predicate. `compileGlobMatchers` had
four exits out of one loop body (`java:S135`); the per-declaration verdicts moved into
`compileDeclaration`, which returns where each is decided, and the loop now only
concatenates. The whole-or-nothing compile is unchanged — every form is still built into a
local list before any of it is returned.

## Checklist

## Additional Notes

`pom.xml` swaps the explicit `jackson-dataformat-yaml` declaration for `org.yaml:snakeyaml`
(same transitive origin, version-managed by the platform BOM) — the parser is the only place
either was used directly.

Docs updated: the README repository-configuration section gains a pattern table and states the
anchoring rule, the `THRILLHOUSEBOT_REVIEW_IGNORED_FILES` row and the `application.properties`
comment say the same thing for the deployment key, and both note that a comma-separated value
cannot carry a `{a,b}` alternation. The website picks these up through its README include.

Base-branch trigger globs (`ReviewTriggerFilter`) are deliberately untouched: branch names are
not paths, and that table already documents its `*` / `**` behaviour precisely.


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/github/RepoSettingsParser.java:175 — Recursive YAML alias may reach Jackson convertValue and throw an uncatchable Error

Confirmed and fixed in 4aba03a. Both `a: &a {self: *a}` and `&a [*a]` threw `StackOverflowError` out of `parse`. The graph is now walked once with an identity set before `convertValue`, and a document that reaches its own container is refused with a warning naming the file. Red proof: both shapes die with `StackOverflowError` when the guard is disabled.

devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/github/RepoSettingsParser.java:174 — IdentityHashMap cycle guard is removed from the path after each child visit, so diamond-shaped shared nodes are re-walked — fine for depth, but the guard's cost is bounded by document size

Confirmed and fixed in aecddc5. It was worse than bounded-by-document-size: a 24-rung ladder of `[*prev, *prev]` lists (49 alias uses, under the ceiling of 50) ran the path-only walk for 44 s and then `convertValue` died with `OutOfMemoryError`. The walk now memoises each node's expanded size, saturates past `MAX_EXPANDED_NODES` (= the code-point ceiling) and refuses the document; `refusesADiamondAliasLadderThatStaysUnderTheAliasCeiling` pins it under a 10 s preemptive timeout (it timed out on the old code) and `keepsADocumentWhoseSharedAliasesStaySmall` pins the positive case.

devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/github/RepoSettingsParser.java:367 — splitPatternList comma/brace handling treats an unbalanced opening brace as consuming all later separators

Confirmed (`splitPatternList("docs/**, **/*.{js, kept/**")` returned two parts, the second swallowing `kept/**`) and fixed in aecddc5: a pre-pass marks every `{` that no `}` closes as literal, so the commas after it still split. Pinned by `anUnclosedOpeningBraceDoesNotSwallowTheRestOfTheScalarList`.

devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/github/RepoSettingsParser.java:345 — splitPatternList's brace counting treats a closing brace with no matching opener as literal (test) but an opening brace never closed swallows the entire remainder of a multi-value scalar

Same defect as the :367 finding; fixed in aecddc5 by the unmatched-opener pre-pass, with the stray-`}` test still green alongside the new stray-`{` one.

PR #821, opened 2026-09-07
fix(review): classify a patchless file when budgeting is disabled

### LINKED ISSUE #785: Patchless files are packed as reviewed when budgeting is disabled, so the file holds no approval and is never disclosed
`#783` gave patchless files their own disclosure class so the rendered copy stops blaming the review
budget for a file GitHub simply returned no content for. That classification lives in
`sizeWithinBudget`, which the disabled-budgeting path never calls.

`DiffBudgetPlanner.renderAndSize`:

```java
for (var file : ordered) {
  var section = formatter.formatFileSection(file, reviewableNames);
  if (diffBudgetTokens <= 0) {
    // The disabled-budgeting path never reads the estimates; skip the BPE pass entirely.
    rendered.sized().add(new Sized(file, section, 0));
  } else {
    sizeWithinBudget(file, section, diffBudgetTokens, rendered);   // <- the patchless check
  }
}
```

So with budgeting off, a patchless file is packed as a `Sized` section of 0 tokens and
`rendered.patchless()` stays empty. `sizeWithinBudget`'s own comment says exactly why that is wrong:

> Its rendered section is a bare header with no ```diff``` body — there is nothing for the model to
> read — so packing it would count the file as fully reviewed and let an unbacked "resolved" claim
> through. Never packed, holds APPROVE, disclosed.

On this path it is packed, so none of those three properties hold: the file is not disclosed, it does
not hold APPROVE, and the bot can approve a pull request containing a binary or too-large-to-render
file it never read.

## Reachability

Not hypothetical, and not a test-only shape. `plan(reviewable, inputs)` routes here whenever the
active model reports no input limit:

```java
if (activeModel.maxInputTokens() <= 0) {
  return plan(reviewable, 0, 1);
}
```

and the method's own javadoc documents that as a supported configuration — *"An explicit
`max-input-tokens <= 0` disables budgeting"*. Any deployment running a model with no configured input
limit is on this path for every review.

## Not a regression, but it undercuts what the release says

The disabled-budgeting branch predates #783, so v0.6.6 does not make this worse. It matters because
of what the release claims: the changelog entry for #783 says patchless files are now disclosed under
their own reason, and on this configuration they are disclosed under no reason at all. The same is
true of #628's original complaint — the wrong-blame wording was fixed for the budgeted path only.

## Direction

Hoist the patchless check above the branch, so classification does not depend on whether budgeting is
enabled — it is a property of the file, not of the budget. The `Sized` packing below can keep its
fast path for everything else.

Worth checking the same way: `unclippable` is also only recorded inside `sizeWithinBudget`, though a
disabled budget cannot produce an over-budget clip, so that one looks sound.

Found while auditing the review threads on #783 against `main`.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [ ] ✨ Feature
- [ ] 📝 Documentation
- [ ] 🔧 Refactor
- [ ] 🚀 Performance
- [ ] ✅ Test
- [ ] 🔒 Security
- [ ] 📦 Dependency update
- [ ] 🏗️ CI/CD

## Description

#783 gave a file GitHub returned no patch text for (a binary, or a text diff too large to display) its own disclosure class, so the review names it under an honest reason and holds APPROVE over it. That classification lived in `sizeWithinBudget`, and the disabled-budgeting branch of `renderAndSize` never calls it: with `max-input-tokens=0` (which is what `plan(reviewable, inputs)` routes to whenever the active model reports no input limit) the file was packed as a 0-token `Sized` section and counted as reviewed.

Verifying the issue against the code showed the gap is wider than the one branch it names. Hoisting the check alone would have recorded the file in `rendered.patchless()` and changed nothing the reader sees, because two more places on the same lane discard the class:

- `plan(reviewable, diffBudgetTokens, maxBatches)` builds the unbudgeted `BudgetPlan` through the convenience constructor that predates the patchless class, so the plan's `patchlessFiles()` stayed empty regardless of what `renderAndSize` recorded.
- `VerdictBuilder.build` builds an empty `TruncationDetail` whenever `plan.budgeted()` is false and takes its omitted count from the legacy line cap alone, so even a plan carrying the class would not have held approval or named the file.

The fix covers all three:

- `renderAndSize` classifies a patchless file ahead of the budget branch. Whether a file has content to review is a property of the file, not of the budget. `sizeWithinBudget` loses its copy of the check.
- The unbudgeted plan is built through the canonical constructor and carries `rendered.patchless()`. It still has no omitted or clipped files: without a budget nothing can exceed one, which is also why `unclippable` is sound as it stands (the second half of the issue). `pack()` and `clipToBudget` never run on that lane.
- The verdict's legacy lane carries `plan.patchlessFiles()` in its `TruncationDetail` and adds the class to the line-cap count. The two are not the same account: the line cap counts sections it dropped past `max-diff-lines`, and a file with no content is a gap whether or not its two-line bare header fit. The existing "never sum both" rule is about plan omissions versus line-cap omissions, which do describe the same files on different lanes, and that rule is unchanged.

One consequence needed its own handling. `coverageGapClause` and `coverageGapBrief` drop the numeric "N file(s) were omitted" wording as soon as the detail names any file, on the reasoning that the count was then already accounted for per class. On the legacy lane that is no longer true: the detail names the patchless files but knows the line-cap omissions only as a count, so a review that hit the line cap and also carried a binary would have named the binary and lost the sentence about the other files. The #659 guard (`!hasFileGaps() && omittedFiles > 0`) is generalized to the remainder past the named classes: `TruncationDetail.namedFileGaps()` sums the six lists, and the numeric clause renders for `omittedFiles - namedFileGaps()` when that is positive. On the budgeted lane the count is built from the same lists the detail carries, so the remainder is zero and nothing changes there; the existing test pinning that (`coverageGapClauseDoesNotAddTheLegacyCountWhenFileGapsAreNamed`) still passes unmodified.

On the issue's claim that this is not a regression: confirmed. The `diffBudgetTokens <= 0` branch predates #783, and the unbudgeted `BudgetPlan` and the legacy verdict lane never carried any file class. The 0.6.6 changelog entry for #783 already scoped itself to the budgeted path and pointed here.

Not changed: the model-facing side of the legacy lane. The single uncapped call sends `ctx.diff()`, which renders a patchless file as a bare `### name (modified, +N -M)` header, and that lane has no withheld-material notice or summary call to name the file to the model. The issue is about the reader-facing disclosure and the approval hold, which this covers; the prompt-side note is a separate change if it is wanted.

## Related Issues

Fixes #785

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

Red first. Five tests written against the unfixed code, failing as:

```
DiffBudgetPlannerTest.configDisabledBudgetingStillClassifiesAPatchlessFile:580 expected: <[assets/logo.png]> but was: <[]>
VerdictBuilderTest.aPatchlessFileHoldsApprovalAndIsDisclosedWhenBudgetingIsDisabled:1142 expected: <[assets/logo.png]> but was: <[]>
VerdictBuilderTest.disabledBudgetingKeepsTheLineCapCountAlongsideAPatchlessFile:1098 expected: <3> but was: <2>
ReviewResultTest.coverageGapClauseKeepsTheLegacyCountForTheFilesTheDetailDoesNotName:811 1 file(s) could not be reviewed because GitHub provided no diff content for them — binary files, or text diffs too large to display (assets/logo.png) ==> expected: <true> but was: <false>
ReviewResultTest.coverageGapBriefKeepsTheLegacyCountForTheFilesTheDetailDoesNotName:847 expected: <2 file(s) omitted, 1 file(s) without diff content from GitHub> but was: <1 file(s) without diff content from GitHub>
```

- `configDisabledBudgetingStillClassifiesAPatchlessFile` goes through `plan(reviewable, inputs)` with `max-input-tokens=0`, so `activeModel.maxInputTokens()` is 0, and asserts the file lands in `patchlessFiles()`, is not packed, is not a budget omission, and makes the plan truncated.
- `aPatchlessFileHoldsApprovalAndIsDisclosedWhenBudgetingIsDisabled` plans through a real `DiffBudgetPlanner` with budgeting off and builds the verdict from that plan: state is COMMENT, `omittedFiles()` is 1, the summary carries the "no diff content" clause naming the file with no budget wording, and the check-run summary carries the "without diff content from GitHub" label.
- `disabledBudgetingKeepsTheLineCapCountAlongsideAPatchlessFile` pins the coexistence: a line-cap count of 2 plus one patchless file gives a count of 3, a held verdict, and both sentences on the banner and both labels on the check-run brief.
- The two `ReviewResultTest` cases pin the remainder rendering in the clause and the brief directly.

Gates, from the worktree root with JDK 25:

- `./mvnw -B clean compile spotbugs:check spotless:check`: `BugInstance size is 0`, BUILD SUCCESS
- `./mvnw -B clean test`: `Tests run: 3636, Failures: 0, Errors: 0, Skipped: 0`, BUILD SUCCESS
- Patch coverage over the `src/main` diff against `origin/main` (JaCoCo, lines and branches): `GAPS: none`

## Checklist

## Screenshots / Logs

N/A

## Additional Notes

The CHANGELOG entry sits under `[Unreleased]` → `Fixed`. No configuration change; a deployment running with `max-input-tokens=0` will see a binary or oversized file in a pull request hold approval and be named in the review banner and the check-run summary, which is what the budgeted configuration has done since 0.6.6.

A corner worth recording rather than handling: on the legacy lane with the line cap on, a patchless file whose bare header itself fell past `max-diff-lines` is inside the line-cap count and also named as patchless, so the banner mentions it twice, once by name and once inside the number. The line cap reports no names, so the overlap cannot be subtracted; the error is in the safe direction (a count one too high on a review that is held either way), and it takes a diff over 5000 lines with the budget disabled to reach it.


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewResult.java:772 — Legacy count can be double-counted with the patchless class

Checked the arithmetic rather than the reading of it: the leading `omittedFileNames().size()` re-adds exactly the names `namedFileGaps()` subtracted, so the named omissions count once and only the unnamed remainder is added. The `Math.max(0, …)` is what matters, and the suggested `omittedFiles - (namedFileGaps() - omittedFileNames.size())` drops it: with two named omissions and two clipped files against an omitted count of 2, it reports 0 omitted files where the detail names two. b261ea2 pins that case (`coverageGapBriefKeepsTheNamedOmissionsWhenClippedFilesOutnumberTheCount` expects `2 file(s) omitted, 2 file(s) partially analyzed`).

@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/ReviewResult.java:315 — namedFileGaps() javadoc misnames the class as 'six classes'

"Classes" is the codebase's name for the six coverage-gap kinds (`hasFileGaps` uses it the same way). The part that was worth fixing is taken in b261ea2: the javadoc now says the sum includes the omitted names, that the callers render only the difference, and that the difference is clamped at zero.

PR #822, opened 2026-09-07
fix(release): keep pre-release and floating tags out of the :latest gate

### LINKED ISSUE #733: The update_latest gate can silently skip the docs deploy instead of failing loudly
Found by deep-audit round 6 against the shipped v0.6.3. Introduced by #717 (rework), which deliberately gates the docs dispatch on `update_latest` and states the job fails loudly when the dispatch is refused.

The loud failure covers a refused dispatch, but not a gate that evaluates false: `sort -V` ranks pre-release and floating tags above a plain release — `0.6.3-rc1` and `nightly` both sort above `0.6.3` — so a repository carrying either would evaluate `update_latest` false for a legitimate release, and the job is **skipped**, not failed. That is the silent miss #717 exists to end, in a different place: nothing on the release page or in the run would say the docs did not publish.

Not currently triggerable (the repo has no pre-release or floating tags), hence latent.

Fix direction: exclude pre-release and non-semver tags from the comparison, and make a false gate on a non-prerelease tag an explicit, logged decision rather than a silent skip.

Verified while checking this: the v0.6.3 release did dispatch and deploy — release run `31915551698` job `publish-docs` success, dispatched docs run `31915660729` (`event=workflow_dispatch`, `head_branch=v0.6.3`) success. First self-published docs deploy in the repository's history.

Proof: audit report AUDIT6-A (finding A6).

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [x] 📝 Documentation
- [x] 🏗️ CI/CD

## Description

The claim in #733 holds. In `release.yml`'s `promote` step the highest existing release is picked with `gh release list ... | sed 's/^v//' | sort -V | tail -1`, and under `sort -V` (GNU coreutils on the runner, and BSD sort agrees) `0.6.8-rc1` and `nightly` both rank above `0.6.8`. A repository carrying a release on either kind of tag would therefore read a legitimate `v0.6.8` as older than the highest release, set `update_latest=false`, and `publish-docs` and `bump-version` would be skipped. A skipped job says nothing, and the only trace was a plain `echo` in the step log. That is the silent miss #717 set out to end, one job earlier than the dispatch it made loud. The repository has no such tags today, so the defect is latent.

Two changes to the `promote` step, and nothing else in the flow:

**Only `vMAJOR.MINOR.PATCH` releases take part in the comparison.** The tag list is filtered with `sed -nE 's/^v([0-9]+\.[0-9]+\.[0-9]+)$/\1/p'`, which replaces the old `sed 's/^v//'` and folds the shape check into it. `sed` rather than `grep` because the step runs under `bash -e -o pipefail`, and a `grep` that matches nothing (a first release, or a repository whose only releases are release candidates) would fail the step; `sed` exits 0 on no match and `HIGHEST_VER` comes back empty, which the existing `-n` guard already handles. Drafts are excluded with `--exclude-drafts`, since a draft is not a release. The `verify` job already holds the incoming tag to that shape, so the new version always participates. A GitHub release *flagged* pre-release but carrying a plain `vX.Y.Z` tag still participates; the rule is the tag, as the issue asks, and this repository's pipeline creates no such release.

**The decision is recorded either way.** After the gate, one `DECISION` line says whether the tag is the highest release and what it was before, or that it is a release on an older line and which version it lost to (with a pointer to `docs/RELEASING.md` for republishing the docs by hand). It goes to `$GITHUB_STEP_SUMMARY` under a `## Promotion` heading and to a `::notice title=Promotion::` annotation, so it shows on the run's summary page and in the annotations list, not only in an expanded step log. The old `echo` is replaced by that line.

What #717 established is preserved: `publish-docs` still runs only when `update_latest == 'true'`, still waits for the run it dispatched, and still fails the job when the dispatch produces no run or the run fails. This PR changes neither that job nor `bump-version`; it only fixes what feeds their gate and makes the gate's outcome visible.

`docs/RELEASING.md` now states which releases count as the highest and where the decision is recorded, and `CHANGELOG.md` has the entry under `[Unreleased]` → `Fixed`.

## Related Issues

Fixes #733
Related: #717 (introduced the gate and the loud failure on a refused dispatch)

## How Has This Been Tested?

- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

**Harness.** The `run:` body of the `Promote image tags` step was extracted verbatim from `release.yml` (from `origin/main` and from this branch), the two `${{ needs.verify.outputs.* }}` expressions substituted with the tag under test and a fixed SHA, and the body executed under `bash -eo pipefail` with `gh` stubbed to print a sample release list and `crane` stubbed to echo its arguments. `GITHUB_OUTPUT` and `GITHUB_STEP_SUMMARY` point at temp files and are printed after each run.

The red run is the step from `origin/main` over `v0.6.7 v0.6.8-rc1 nightly v0.6.6` with `v0.6.8` as the new tag: it picks `nightly` as the highest release and sets `update_latest=false`, with an empty step summary. The same list against this branch promotes `v0.6.8` and records the decision; the remaining runs cover a patch on an older line (`v0.6.5`, held back and logged with the version it lost to), a repository with only release candidates and a floating tag, a first release with no prior releases at all (the `pipefail` case), and a plain list.

```
===== origin/main promote step (red) =====
--- NEW=v0.6.8 releases=[v0.6.7 v0.6.8-rc1 nightly v0.6.6]
    crane tag img:abc1234 v0.6.8
    crane tag img:abc1234-distroless v0.6.8-distroless
  Skipping :latest — v0.6.8 is older than highest release vnightly
  rc=0  GITHUB_OUTPUT: update_latest=false  STEP_SUMMARY: 

===== this branch's promote step (green) =====
--- NEW=v0.6.8 releases=[v0.6.7 v0.6.8-rc1 nightly v0.6.6]
    crane tag img:abc1234 v0.6.8
    crane tag img:abc1234-distroless v0.6.8-distroless
    crane tag img:abc1234 latest
    crane tag img:abc1234-distroless latest-distroless
  ::notice title=Promotion::v0.6.8 is the highest release (previously v0.6.7): :latest, the docs site and the version bump follow it
  rc=0  GITHUB_OUTPUT: update_latest=true  STEP_SUMMARY: ## Promotion  v0.6.8 is the highest release (previously v0.6.7): :latest, the docs site and the version bump follow it 
--- NEW=v0.6.5 releases=[v0.6.7 v0.6.8-rc1 nightly v0.6.6]
    crane tag img:abc1234 v0.6.5
    crane tag img:abc1234-distroless v0.6.5-distroless
  ::notice title=Promotion::v0.6.5 is a release on an older line, below v0.6.7: :latest, the docs site and the version bump stay where they are (see docs/RELEASING.md to republish the docs by hand)
  rc=0  GITHUB_OUTPUT: update_latest=false  STEP_SUMMARY: ## Promotion  v0.6.5 is a release on an older line, below v0.6.7: :latest, the docs site and the version bump stay where they are (see docs/RELEASING.md to republish the docs by hand) 
--- NEW=v0.6.8 releases=[v0.6.8-rc1 v0.6.8-rc2 nightly]
    crane tag img:abc1234 v0.6.8
    crane tag img:abc1234-distroless v0.6.8-distroless
    crane tag img:abc1234 latest
    crane tag img:abc1234-distroless latest-distroless
  ::notice title=Promotion::v0.6.8 is the highest release: :latest, the docs site and the version bump follow it
  rc=0  GITHUB_OUTPUT: update_latest=true  STEP_SUMMARY: ## Promotion  v0.6.8 is the highest release: :latest, the docs site and the version bump follow it 
--- NEW=v0.1.0 releases=[]
    crane tag img:abc1234 v0.1.0
    crane tag img:abc1234-distroless v0.1.0-distroless
    crane tag img:abc1234 latest
    crane tag img:abc1234-distroless latest-distroless
  ::notice title=Promotion::v0.1.0 is the highest release: :latest, the docs site and the version bump follow it
  rc=0  GITHUB_OUTPUT: update_latest=true  STEP_SUMMARY: ## Promotion  v0.1.0 is the highest release: :latest, the docs site and the version bump follow it 
--- NEW=v0.6.8 releases=[v0.6.7 v0.6.6 v0.6.5]
    crane tag img:abc1234 v0.6.8
    crane tag img:abc1234-distroless v0.6.8-distroless
    crane tag img:abc1234 latest
    crane tag img:abc1234-distroless latest-distroless
  ::notice title=Promotion::v0.6.8 is the highest release (previously v0.6.7): :latest, the docs site and the version bump follow it
  rc=0  GITHUB_OUTPUT: update_latest=true  STEP_SUMMARY: ## Promotion  v0.6.8 is the highest release (previously v0.6.7): :latest, the docs site and the version bump follow it
```

**actionlint.** `actionlint -color` (1.7.12, shellcheck enabled, with the `.actions-ref` side-path mirrored the way the CI job does it) exits 0 on the tree.

The `publish-docs` job itself can only be proven end to end by the next release, as #717 noted; its code is unchanged here.

No Java changes; the Java gates do not apply.

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works

## Screenshots / Logs

See the harness output above.

## Additional Notes

Branched from `origin/main`, independent of #818. Both add a subsection under `[Unreleased]` in `CHANGELOG.md` (`Changed` there, `Fixed` here), so whichever merges second will need a trivial conflict resolution in that file.


devops-thiago:
@thrillhousebot resolved .github/workflows/release.yml:190 — gh release list --limit 200 may truncate releases, affecting highest-release comparison

Taken in 4784a6a: the listing asks for up to 1000 releases (`gh release list` pages through them). The repository has 16 releases today, so 200 was not close to binding, but the higher ceiling costs nothing and removes the question.

PR #826, opened 2026-09-07
fix(review): size the prompt overhead from a fixed-width fence

### LINKED ISSUE #586: Disclosure classification races under the parallel batch pass (flaky test, possible real bug)
`ChangelogEntryGeneratorTest.discloseTheShortfallWhenOneBatchFails` failed once in CI on PR #585 and does not reproduce locally — 5/5 green for the class, and the full suite green at the same 2769-test count CI ran, on the same merge. So it is intermittent, and unrelated to that PR's change (disabling chat memory).

The assertion output shows the posted body disclosed the files as **clipped**, not as uncovered:

```
> ⚠️ **Large PR — partial coverage.** 2 file(s) were only partially analyzed
  (src/Foo.java, src/Other.java) because the diff exceeded the review budget
```

The test is about a *failed batch*, whose files should be disclosed as not-reviewed. Getting the "partially analyzed" wording instead means a different coverage class won for those files on that run.

Batches are processed in a parallel pass (`FindingPipeline` fans out over a virtual-thread executor), and the coverage classes are recorded onto the **shared** `BudgetPlan` from those concurrent tasks — `recordUncoveredFiles`, `recordClippedFiles`, `recordResponseCutFiles`. The de-duplication that decides which class a file ends up in (`effectiveOmittedFiles` / `effectiveClippedFiles` / `withoutNames` in `VerdictBuilder`) reads those lists after the fact. If the ordering of those recordings varies, the class a file is disclosed under can vary too.

So this may be more than a flaky test: the same race would make the **user-visible disclosure** nondeterministic on a real review — a file that was never reviewed could be reported as merely partially analyzed, which understates the gap and is exactly the kind of misstatement the disclosure classes exist to prevent.

Worth establishing first whether the nondeterminism is in the production recording path or only in the test's mock ordering. If it is production, the fix is to make the classification deterministic regardless of completion order rather than to relax the test. Reproduction will likely need repetition under load — it survived 5 local runs and a full suite.

### LINKED ISSUE #604: A fresh random fence per call makes budget plans non-reproducible run to run
Found while diagnosing #586.

`PromptTemplateEscaper.fence()` mints a fresh 32-hex-char CSPRNG token per call. A random hex string does not tokenize to a fixed BPE width — measured over 2000 draws: **min 53, max 81 tokens, spread 28**.

Both `AbstractPrSuggestionGenerator.sharedPromptOverhead` (shared by `/describe`, `/improve`, `/add-docs`, `/generate-tests`) and `DiffBudgetPlanner.plan(...)` size their overhead by drawing a **live** fence. So the overhead subtracted from the budget varies by up to ~30 tokens between runs of the same input, and the resulting batch plan is not reproducible.

**This is not a production bug.** The sample is an unbiased estimate of an overhead that genuinely is random per call, and a ~30-token error is absorbed by `token-safety-margin` at real budgets of 10⁵. It is recorded because:

1. It cost real time to diagnose. In #586 it was the entire cause of a 1-in-1500 CI failure that had been filed as a possible concurrency race in the review pipeline — a hypothesis it took a full investigation to refute.
2. Sibling tests that set a small `max-input-tokens` carry the same latent fragility, and will fail the same way, rarely, for the same reason. #586's fix pads its own fixtures past the jitter but does not address the others.
3. Non-reproducible planning is awkward to reason about generally — the same PR can batch differently on two runs with no input change.

Suggested direction: size the overhead from a fixed-width representative fence (the width is what matters, not the value; the real fence still gets used in the emitted prompt). Plans then become reproducible for a given input, and the class of rare test flake disappears rather than being padded around.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [ ] ✨ Feature
- [ ] 📝 Documentation
- [ ] 🔧 Refactor
- [ ] 🚀 Performance
- [ ] ✅ Test
- [ ] 🔒 Security
- [ ] 📦 Dependency update
- [ ] 🏗️ CI/CD

## Description

`PromptTemplateEscaper.fence()` mints a fresh 32-hex-character CSPRNG token per call, and a random hex string has no fixed BPE width. Both places that size the shared prompt overhead, `DiffBudgetPlanner.plan(reviewable, inputs)` for the review and `AbstractPrSuggestionGenerator.sharedPromptOverhead` for `/describe`, `/improve`, `/add-docs` and `/generate-tests`, called `fence(" ")` live. The overhead the planner subtracts from the per-call budget was therefore a random variable, the diff budget for one input landed in a window tens of tokens wide, and the same pull request could batch differently on two runs. That is the whole cause of the flake the issue's comments trace: #586 in `ChangelogEntryGeneratorTest`, then `DocGenerationServiceTest.keepsTheDocsFromTheBatchesThatSucceededWhenOneBatchFails` on #596, a commit that changed only javadoc text.

Every claim in the issue checks out against the code, with one number to update. The issue measured 53 to 81 tokens over 2000 draws; over 200,000 draws with the project's own tokenizer (jtokkit 1.1.0, `cl100k_base`) the two fence lines around a space run 51 to 87, so the real window is 36 tokens, wider than the sample suggested. The histogram, for the record:

```
51=2 53=103 55=895 57=4149 59=11978 61=23650 63=35083 65=39023 67=34895 69=24821
71=14448 73=6950 75=2828 77=879 79=238 81=44 83=10 85=3 87=1
```

The fix follows the issue's direction: the overhead is sized from a fixed-width representative fence while the emitted prompt still gets a fresh random one. `PromptTemplateEscaper.fenceForBudgeting()` returns the two fence lines around a single space, shaped exactly as `fence(" ")` emits them, with the token `1a2b3c4d5e6f1a2b3c4d5e6f1a2b3c4d` in place of the random one. Both overhead callers use it; `fence()` itself is unchanged apart from sharing the line-building code.

The representative is chosen so that it is the widest a draw can be, which makes the bound a property of the tokenizer rather than of the sample. `cl100k_base`'s pre-tokenizer cuts a run of letters, or a run of up to three digits, into one chunk, and BPE merges only inside a chunk. A token that alternates digit and letter is therefore 32 single-character chunks, one token each: 46 tokens a line, 93 for the stand-in. A chunk never costs more tokens than it has characters, and a random token's longer runs can only merge further, so no draw can cost more than 93. The measured maximum of 87 sits under it, and the estimate errs by at most the difference, on the side of a smaller batch, which is the direction `token-safety-margin` already errs in. The measurement and the argument are recorded on `WIDEST_FENCE_TOKEN`.

The sibling test helpers that computed their `max-input-tokens` from a live `fence(" ")` (`ChangelogEntryGeneratorTest`, `DocGenerationServiceTest`, `PrDescriptionGeneratorTest`, `PrImprovementServiceTest`, `UnitTestGeneratorTest` in two places) now use the same stand-in, so the room a `budgetWithDiffRoom(n)` call requests is what the planner gets, to the token. That is what makes all eleven `budgetWithDiffRoom(40)` sites in `DocGenerationServiceTest` deterministic at once rather than padding each one.

## Related Issues

Fixes #604

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

Red first. Two tests written against the unfixed code, failing as:

```
DiffBudgetPlannerTest.planningTheSameInputRepeatedlyGivesTheSameBatches:620 one input, one plan: [1 batch(es), covered [dir/f1.java, dir/f2.java], clipped [], 2 batch(es), covered [dir/f1.java, dir/f2.java], clipped []] ==> expected: <1> but was: <2>
AbstractPrSuggestionGeneratorTest.sharedPromptOverheadIsTheSameStringForTheSameInputs:47 expected: <systemuser[[THRILLHOUSEBOT-UNTRUSTED-DATA-67fad00e11011ce5658189bfa9c0e614]] …> but was: <systemuser[[THRILLHOUSEBOT-UNTRUSTED-DATA-<a different token>]] …>
```

- `planningTheSameInputRepeatedlyGivesTheSameBatches` goes through `plan(reviewable, inputs)` with a budget set 30 tokens under the two fixture sections, inside the window a live draw swept, and plans the same input 200 times. Against the unfixed code the plans split between one batch and two, as the failure line shows; the fixed overhead lands every plan on two batches, which the test now also pins. The red run built the stand-in fence locally in the test so it compiled against the unfixed code; the fix commit replaces that local helper with `fenceForBudgeting()`, and the arithmetic and assertions are unchanged.
- `sharedPromptOverheadIsTheSameStringForTheSameInputs` calls the generator's overhead twice for one input and asserts the strings are equal; before the fix they differ in the fence token.
- `fenceForBudgetingIsFixedAndShapedLikeARealFence` and `fenceForBudgetingIsAtLeastAsWideAsAnyRealFence` in `PromptTemplateEscaperTest` pin the stand-in: the same string on every call, a real fence's shape and character length, exactly 93 tokens, and at least as wide as each of 2000 fresh draws, so a change to the token or the tokenizer that broke the bound fails there rather than as a flake elsewhere.

Gates, from the worktree root with JDK 25:

- `./mvnw -B clean compile spotbugs:check spotless:check`: `BugInstance size is 0`, BUILD SUCCESS
- `./mvnw -B clean test`: `Tests run: 3635, Failures: 0, Errors: 0, Skipped: 0`, BUILD SUCCESS
- Patch coverage over the `src/main` diff against `origin/main` (JaCoCo, lines and branches): `GAPS: none`

## Checklist

## Screenshots / Logs

N/A

## Additional Notes

On the #601 padding. #601 fixed #586 by padding `ChangelogEntryGeneratorTest`'s fixtures with 30 filler lines each (`PAD_LINES`), raising `ROOM_FOR_ONE_FILE` to 400, and adding `batchingFixturesLeaveMarginWiderThanTheFenceJitter`, which pins every batching margin above a 64-token bound on the jitter. With both sides sizing from the same fixed string the jitter is zero and the padding is slack rather than a requirement, so it can be undone: `PAD_LINES` back to 0, `ROOM_FOR_ONE_FILE` re-derived against the unpadded sections, and the pin test either dropped or kept with a zero bound. That is not a trivial revert, since the room constant has to be re-derived and every batching assertion in the class re-checked against the new shape, so this PR leaves the padding in place and only rewrites the prose on `PAD_LINES`, `FENCE_JITTER_TOKENS` and `budgetFor` so it stops describing the two-draw jitter as current. Keeping the pin test has some value on its own: it would fail loudly if a live draw ever crept back into either overhead. A follow-up can take the padding out if the smaller fixtures are preferred.

Not changed: `fence()`'s security property. The random token still guards the emitted prompt; the stand-in never reaches a model, and `fenceUsesAFreshTokenEachCall` still holds.


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/PromptTemplateEscaper.java:36 — Comment claims a 93-token bound that the code does not enforce

Fair on the wording, taken in 0d6696e: the javadoc now says the bound holds under cl100k_base as jtokkit 1.1.0 ships it, that it belongs to the tokenizer rather than to this class, and that this is why the test pins the number.

@thrillhousebot resolved src/test/java/dev/thiagogonzaga/thrillhousebot/review/PromptTemplateEscaperTest.java:130 — Test asserts a hard 93-token bound that is not guaranteed by the code

Pinning the tokenizer's behaviour is the point of that assertion. The overhead is sized from the stand-in on the assumption that no real fence is wider; if a jtokkit update moved the stand-in's width, the budgeting would quietly change, and #604 was a rare flake that took a full investigation to trace. A failing test at the upgrade is the outcome we want.

@thrillhousebot resolved src/test/java/dev/thiagogonzaga/thrillhousebot/review/DiffBudgetPlannerTest.java:600 — Test's 200-iteration loop may not be sufficient to prove determinism

Agreed that 200 plans is a regression guard with limited power against a live fence, which is why it is not the only one: `AbstractPrSuggestionGeneratorTest.sharedPromptOverheadIsTheSameStringForTheSameInputs` compares the overhead string itself across two calls, so a live fence reintroduced into the overhead fails deterministically on the first run.

Issue #784, opened 2026-08-24
Throttle and content-creation classification are word-order whitelists, so a reworded 403 silently loses the retry budget


devops-thiago:
`GitHubApiError` decides two things from the wording of a 403 body: whether the failure is a throttle
worth repeating at all (`THROTTLE_WORDING`), and whether it is the content-creation block whose
measured 72-second width the retry budget is sized against (`CONTENT_CREATION_BLOCK`). Both are fixed
alternations of specific phrases:

```java
THROTTLE_WORDING        = secondary rate limit | abuse detection | rate limit exceeded
                          | blocked from (content creation | creating content)
CONTENT_CREATION_BLOCK  = abuse detection | blocked from (content creation | creating content)
```

A block message GitHub words differently matches neither. The consequences are not symmetric with the
usual "we miss a case" defect:

- missing `CONTENT_CREATION_BLOCK` drops the 30-second floor, so the derived delay falls back to the
  linear 5s/10s/15s and the whole budget expires inside a window three times as wide — the exact
  failure #722 was filed on;
- missing `THROTTLE_WORDING` is worse: the call is not read as a throttle at all, so it fails fast
  with **no repeat**, and the generated content is lost on the first refusal.

Both are silent. Nothing logs "this looked like a throttle but matched no pattern".

## Why it is being filed rather than widened again

#722 already widened these twice under review, each time for a variant someone imagined: first
`blocked from content creation`, then the gerund `blocked from creating content`. That is the shape of
a whitelist being patched one guess at a time, and the guesses came from a reviewer rather than from
GitHub's documented set. Widening again on the same basis buys nothing durable.

Two directions worth weighing, neither obviously right:

1. **Establish the real set.** GitHub documents its secondary-limit messages; collecting the actual
   variants for content creation and matching those is a bounded, checkable fix. The measured body
   is `You have exceeded a secondary rate limit and have been temporarily blocked from content
   creation.`
2. **Stop deciding from prose.** The status plus the absence of primary-quota exhaustion
   (`x-ratelimit-remaining` well above zero on a 403) already identifies a secondary limit without
   reading the message at all. That is how the measured incident presented: `403` with
   `remaining=4771`. Wording could then narrow the class rather than gate it.

Deliberately not proposed: matching the bare noun phrase `content creation`. This regex runs over an
attacker-influenced response body on its way to an operator's log, and the loose phrase can appear in
a message that is not this block.

## Instrumentation worth having either way

[...]

PR #834, opened 2026-09-07
fix(github): ship the write pacer's ceiling at the budget it is pinned to

### LINKED ISSUE #830: Write pacer ceiling: the 90 s constant pinned to the retry budget is shadowed by the 60 s property default
Two loose ends noticed while working #784 and #734, neither a defect on its own.

## 1. The pacer's effective ceiling is 60 s, and the test that pins it to the retry budget guards a value production never uses

`GitHubWritePacer.DEFAULT_MAX_WAIT` is 90 s, and `GitHubWritePacerTest` pins it equal to `GitHubWriteRetry.TOTAL_BUDGET` (the #723 reasoning: a caller waiting for its content-creation slot should not give up before the retry that follows would have). But `application.properties` ships

```
thrillhousebot.github.write-max-wait=${GITHUB_WRITE_MAX_WAIT:60s}
```

and the pacer reads `configured(MAX_WAIT_KEY, DEFAULT_MAX_WAIT)`, so the constant only applies when the property is absent, which it never is in a packaged build. The README documents `60s`. Either the property default should be `90s` (and the README with it) so the pin means what it says, or the pin should read the configured value and the 90 s constant should go. Decide which, and make the javadoc on `DEFAULT_MAX_WAIT` say where the effective value comes from.

## 2. A generic secondary limit with neither header waits 5/10/15 s where GitHub asks for at least 60 s

GitHub's secondary-limit guidance (and octokit's throttling plugin) wait a minimum of 60 s when a 403 names a secondary limit and carries neither `Retry-After` nor a reset header. `GitHubWriteRetry.retryDelay` keeps the linear 5/10/15 s for that case; the 30 s floor applies only to the content-creation block. That is a deliberate trade recorded on #722/#723 (the measured block width was 72 s, and the total budget is 90 s), and #784 left it alone. Recording it here so it is a decision, not an omission: if the generic case is ever observed in the logs (the new WARN from #784 will show it), revisit with the measured width.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [x] ✅ Test

## Description

`GitHubWritePacer.DEFAULT_MAX_WAIT` is 90 s and `GitHubWritePacerTest` pins it to `GitHubWriteRetry.TOTAL_BUDGET` (#723: waiting for a pacing slot must never be worse than being refused and repeated). But the pacer reads `thrillhousebot.github.write-max-wait` first, and that property shipped as `60s`; the constant only applies to a build with no `application.properties`. So the pin guarded a value production never used, and a caller gave up on its slot 30 s before the refusal-and-repeat path would have.

- `application.properties`, `.env.example` and the README default now say `90s`, with the reason next to the property.
- The constant's javadoc states where the effective value comes from.
- A second pin reads the shipped property through SmallRye with Quarkus's `DurationConverter` (the `DevSchemaManagementTest` pattern) and asserts it equals the constant, so the two cannot drift again.

The second half of #830 (5/10/15 s backoff on a generic secondary limit with neither header, where GitHub's guidance is 60 s) is a recorded decision on the issue and needs no code; #784's new WARN will show if the case is ever observed.

## Related Issues

Fixes #830

## How Has This Been Tested?

- [x] Unit tests

Red on unfixed code: `GitHubWritePacerTest.theShippedPropertyDefaultCarriesTheSameCeiling:189 expected: <PT1M30S> but was: <PT1M>`. Green after the property change. `./mvnw -B clean compile spotbugs:check spotless:check` (`BugInstance size is 0`) and the full `./mvnw -B clean test`: 3632 tests, 0 failures. Patch coverage: no uncovered lines (the `src/main` change is javadoc only).

## Checklist

## Screenshots / Logs

N/A

## Additional Notes

Touches `application.properties` and `README.md` near the lines #827 adds for `GITHUB_WRITE_RETRY_BUDGET`; if #827 lands first this rebases cleanly (different lines), and the `[Unreleased]` changelog section will need the usual keep-both resolution.


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/config/ThrillhouseConfig.java:95 — Config interface default still 60s; shipped property change does not change effective default

Right, and fixed in d88852c: `ThrillhouseConfig.GitHubConfig.writeMaxWait` declares `@WithDefault("90s")` now, so the interface, the shipped property and `GitHubWritePacer.DEFAULT_MAX_WAIT` carry the same number.

@thrillhousebot resolved src/test/java/dev/thiagogonzaga/thrillhousebot/github/GitHubWritePacerTest.java:189 — New pin reads application.properties directly, not the config interface default

Also d88852c: `theConfigInterfaceDefaultCarriesTheSameCeiling` reads the `@WithDefault` value off `writeMaxWait` and asserts it equals `DEFAULT_MAX_WAIT`, next to the existing pin on the shipped property, so neither copy can drift from the constant.

PR #837, opened 2026-09-09
fix(github): read the client's 404 on a repository config as no config

### LINKED ISSUE #836: A missing repository config is read as a transient failure: the client's 404 is ClientWebApplicationException, not NotFoundException
`RepoSettingsResolver.fetchAndParse` maps a missing `.github/thrillhousebot.yml` to `ABSENT` only when the client throws `jakarta.ws.rs.NotFoundException`. The RESTEasy Reactive client never throws that: a 404 arrives as `org.jboss.resteasy.reactive.ClientWebApplicationException`, a sibling under `WebApplicationException`, which #481's narrowed catch reads as a transient read failure.

## Observed

Production, v0.6.7, every review of a repository with no config file (this one, for instance):

```
WARN  [RepoSettingsResolver] Could not read repository config .github/thrillhousebot.yml for devops-thiago/ThrillhouseBot: org.jboss.resteasy.reactive.ClientWebApplicationException: Received: 'Not Found, status code 404' when invoking REST Client method: 'GitHubPullRequestClient#getFileContentOnce'
WARN  [RepoSettingsResolver] Could not read repository config .github/thrillhousebot.yaml for devops-thiago/ThrillhouseBot: … status code 404 …
WARN  [RepoSettingsResolver] Could not read a repository config for devops-thiago/ThrillhouseBot; running on the global settings for this review
```

## Effect

- The `ABSENT` branch is dead in production, so "this repository has no config" is never negative-cached: both candidate paths are fetched on every review (two GitHub calls each time) instead of once per minute.
- Three WARN lines per review that describe a transient failure where there was none, on the log level operators watch.
- Behaviour is otherwise correct: the review runs on the global settings either way.

## Why

#481 replaced a catch of `WebApplicationException` (which had also swallowed 5xx and 403 as "no config") with `NotFoundException` for the 404 case and `WebApplicationException | ProcessingException` for everything else. The unit tests stub `NotFoundException`, which is what a JAX-RS server throws, so they passed; the client throws `ClientWebApplicationException` carrying the status on its `Response`.

## Fix

Decide by status, not by subclass: `WebApplicationException` whose `getResponse().getStatus() == 404` is `ABSENT` (keep `NotFoundException` too, for the tests and any client that does throw it); every other status stays `READ_FAILED`. Pin with a test that throws `new ClientWebApplicationException(Response.status(404).build())` from the mocked client and asserts `ABSENT` plus the negative cache, and one with a 503 that asserts `READ_FAILED` and no cache entry.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [x] ✅ Test

## Description

`RepoSettingsResolver.fetchAndParse` treated a missing `.github/thrillhousebot.yml` as `ABSENT` only on `jakarta.ws.rs.NotFoundException`. The RESTEasy Reactive client never throws that; its 404 is `org.jboss.resteasy.reactive.ClientWebApplicationException` with the status on the response, which #481's `WebApplicationException | ProcessingException` catch read as a transient failure. Seen on production (v0.6.7) on every review of a repository with no config:

```
WARN  [RepoSettingsResolver] Could not read repository config .github/thrillhousebot.yml for devops-thiago/ThrillhouseBot: org.jboss.resteasy.reactive.ClientWebApplicationException: Received: 'Not Found, status code 404' …
```

Three WARN lines per review, both candidate paths fetched every time, and the negative cache never written.

The fix decides by status: a `WebApplicationException` whose response is a 404 (and `NotFoundException`, which the tests and any server-side JAX-RS path still throw) is `ABSENT` and cached; every other status stays `READ_FAILED` and is asked again next review; `ProcessingException` unchanged.

## Related Issues

Fixes #836

## How Has This Been Tested?

- [x] Unit tests

Red on unfixed code: `RepoSettingsResolverTest.theClientsOwn404IsNoConfigAndIsCachedLikeOne the client's 404 is a real answer and is cached ==> expected: <1> but was: <0>`. Green after the fix, with `theClientsOwn503IsStillAFailedReadAndIsNotCached` pinning the other direction (the client's 503 stays uncached). `./mvnw -B clean compile spotbugs:check spotless:check` (`BugInstance size is 0`) and the full `./mvnw -B clean test`: 3633 tests, 0 failures. Patch coverage: no uncovered lines or branches.

## Checklist

## Screenshots / Logs

The production log line quoted above, from the bot's own review of #834 at 11:10 UTC today.

## Additional Notes

Adds to `[Unreleased]` in `CHANGELOG.md` like the other open v0.6.8 PRs; the usual keep-both resolution applies.


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/github/RepoSettingsResolver.java:210 — statusOf may NPE if WebApplicationException has no response

Checked against the class the client throws, not assumed. In jakarta.ws.rs-api 4.0.0 every `WebApplicationException` constructor delegates to `WebApplicationException(String, Throwable, Response)`, and that one stores `Response.serverError().build()` when it is handed a null response (`javap -c`: `aload_3; ifnonnull 23; ... Response.serverError().build(); putfield response`). `getResponse()` is not overridden by `ClientWebApplicationException`, the only subclass this catch receives from the REST client. So there is always a response to read, which is what the javadoc says; a null guard would be a branch no input can take, and the patch-coverage gate rejected exactly that branch when it was there.

PR #873, opened 2026-09-16
fix(review): record the shape a dropped file_summaries recovery read

### LINKED ISSUE #872: A dropped file_summaries recovery leaves a blank walkthrough and no record of the shape that arrived
## Background

A review of `kart_backend#509` on 2026-09-16 at 12:54:05 UTC logged:

```
WARN  [ReviewResponseParser] Review response file_summaries did not match the schema — recovered 0 entries and dropped 4; unrecovered entries render as blank walkthrough rows
```

All four entries were dropped, so that review's Changed Files table rendered with no summaries. The stored response for that session (`reviewsession.airesponsejson`, id 4948) holds `summary.file_summaries: []`, because the column is written after normalization: the shape the model actually sent is not recorded anywhere.

Two rounds of the same pull request minutes earlier (ids 4945 and 4946) stored four well-formed `{path, summary}` entries each, so the model produces the conforming shape most of the time and something specific about this response was different.

## The problem

`normalizeFileSummaries` recovers a wide set of shapes (conforming array, non-conforming array, single entry, map form, `"path: summary"` strings) and drops what it cannot read. The warning counts what was dropped but says nothing about what arrived: not the node type, not the field names, not one example. When every entry is dropped, the log gives no way to tell whether the model used an unknown key, emitted non-textual values, nested the entries a level deeper, or sent something else. The response is gone by then, since only the normalized form is stored.

So a walkthrough silently degrades and the cause cannot be diagnosed after the fact, which is the same class of gap #536 and #637 closed elsewhere.

## Change required

- Make the warning diagnosable. When entries are dropped, log the node type of `file_summaries`, and for the first dropped entry its node type and its field names (names only, not values, all of it through `LogSafe` and bounded). Field names are model-supplied, so they are untrusted text and must be treated as such.
- With the shape then known, decide whether `PATH_KEYS`/`SUMMARY_KEYS` should grow, or whether the entry shape needs its own recovery. Do not guess at new keys before the evidence is in: add the logging first, and widen recovery only for a shape actually observed.
- Consider recording the pre-normalization `file_summaries` node on the session row when recovery drops everything, so a maintainer can read it. Weigh that against the row size, and leave it out with a note if the logging alone is enough.

## Constraints

- No change to the conforming path: an array of textual `{path, summary}` entries is still returned untouched.
- Nothing model-supplied reaches a log line without `LogSafe`; `LogSafeInvariantTest` enforces that.
- The recovery stays silent about entries it does recover, so a one-entry miss does not start logging on every review.

## Acceptance criteria

- [ ] A response whose `file_summaries` entries are all dropped logs the node type and the first entry's field names, sanitized and bounded.
- [ ] A response that conforms logs nothing new.
- [ ] Unit tests cover: all entries dropped, some dropped, and the conforming case.
- [ ] `./mvnw -B clean compile spotbugs:check spotless:check` clean (`BugInstance size is 0`) and the full suite green.

## References

- `src/main/java/dev/thiagogonzaga/thrillhousebot/review/ai/ReviewResponseParser.java` (`normalizeFileSummaries`, `addFileSummary`, `conformsToFileSummarySchema`, `PATH_KEYS`, `SUMMARY_KEYS`)
- Production log, dev.noonly, 2026-09-16T12:54:05Z (v0.6.8, glm-5.3-flash); sessions 4945, 4946 and 4948 on `kart_backend#509`.

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [ ] ✨ Feature
- [ ] 📝 Documentation
- [ ] 🔧 Refactor
- [ ] 🚀 Performance
- [ ] ✅ Test
- [ ] 🔒 Security
- [ ] 📦 Dependency update
- [ ] 🏗️ CI/CD

## Description

`ReviewResponseParser.normalizeFileSummaries` recovers `summary.file_summaries` from several shapes the schema does not accept — a map keyed by path, entries keyed `file`/`description`, `"path: summary"` strings, a single object where the array belongs — and drops what it cannot read. The warning it ends on counted the drops and said nothing about what it had read:

```
Review response file_summaries did not match the schema — recovered 0 entries and dropped 4; unrecovered entries render as blank walkthrough rows
```

That is the whole record of a review whose Changed Files table rendered with no summaries. The session row stores the response after this rewrite, so `summary.file_summaries` reads `[]` there, and by the time anyone looks the shape that arrived is gone. A count cannot tell an unknown key from a non-textual value, from a level of nesting nobody expected, or from a scalar.

When the recovery saves nothing, the warning now also carries the node type `file_summaries` arrived as, and the node type and field names of the first dropped entry:

```
Review response file_summaries did not match the schema — recovered nothing from 2 entries, so the walkthrough renders with no summaries; it arrived as ARRAY, first entry OBJECT with field(s) [file_path, change_summary]
```

Names only, never values: a name is all it takes to decide whether `PATH_KEYS`/`SUMMARY_KEYS` should grow, while a value carries the model's prose about the diff. The model chooses those names, so each one goes through `LogSafe.oneLine` at the point it is read, and the line is bounded in both directions — at most eight names, each cut at forty characters, with the rest counted (`… and 3 more`) — since an entry is free to carry a thousand keys or one key a megabyte long.

Three things deliberately do not change:

- **The conforming path.** An array of textual `{path, summary}` entries still returns before any of this and logs nothing.
- **Partial recovery.** A recovery that saves some entries keeps the count line it always logged and describes no shape, so one stray entry in an otherwise readable array does not start logging a shape on every review.
- **The recovery itself.** No key is added to `PATH_KEYS` or `SUMMARY_KEYS`. The point of the change is to capture the evidence; the production response is not available to read, and `file_path`/`change_summary` in the test is a stand-in chosen to exercise the code, not an observed shape. Widening the lists on a guess would be the same mistake in a new place.

### Verified against the code

Every claim in the issue holds. `normalizeFileSummaries` does drop unreadable entries and log only counts; `PATH_KEYS` and `SUMMARY_KEYS` are as the issue lists them; and the response the session row keeps is written after normalization, so the pre-normalization node is not recoverable from it.

### On recording the node on the session row

Left out, on purpose. `ReviewSession.aiResponseJson` is a single `TEXT` column holding the whole serialized response, so preserving the original `file_summaries` means either a second column and a migration or an ad-hoc field inside that JSON, and what it would hold is a whole walkthrough's worth of model prose — the largest free-text field in the response after the findings — written on exactly the rounds that already failed. The question it would answer is "what shape arrived", and the node type plus the first entry's field names answer that in one bounded log line that costs nothing on a healthy review. The javadoc on `normalizeFileSummaries` states that trade-off so the next reader does not re-open it. If a real case ever needs the values rather than the shape, that is the moment to pay for the column.

## Related Issues

Fixes #872

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

New `FileSummariesDropDiagnosticsTest` captures the `LogRecord` the parser emits, so the assertions do not depend on which handler is installed. Five tests: an all-dropped array, the map form / a scalar / a single entry (each of the three nodes the description can report on), the sanitizing and bounding of the names, a partly recovered response, and a conforming one.

Red, against the unfixed parser:

```
[ERROR] Tests run: 5, Failures: 3, Errors: 0, Skipped: 0
[ERROR]   FileSummariesDropDiagnosticsTest.anAllDroppedRecoveryNamesTheNodeTypeAndTheFirstEntrysFieldNames:110 Review response file_summaries did not match the schema — recovered 0 entries and dropped 2; unrecovered entries render as blank walkthrough rows ==> expected: <true> but was: <false>
[ERROR]   FileSummariesDropDiagnosticsTest.aMapFormAScalarAndASingleEntryAreEachDescribed:118 Review response file_summaries did not match the schema — recovered 0 entries and dropped 1; unrecovered entries render as blank walkthrough rows ==> expected: <true> but was: <false>
[ERROR]   FileSummariesDropDiagnosticsTest.theFieldNamesAreSanitizedAndBoundedInCountAndLength:140 Review response file_summaries did not match the schema — recovered 0 entries and dropped 1; unrecovered entries render as blank walkthrough rows ==> expected: <true> but was: <false>
```

The two guard tests — the partly recovered response and the conforming one — passed before the fix and still pass, which is what makes them guards.

Green, with the fix:

```
[INFO] BugInstance size is 0
[INFO] BUILD SUCCESS

[INFO] Tests run: 4042, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Patch coverage on the `src/main` diff against `origin/main`, line and branch, from `target/site/jacoco/jacoco.xml`:

```
GAPS: none
```

## Checklist

## Additional Notes

The sanitizing test drives the real forgery characters through the parser: a field name carrying U+000A and a bidi override (U+202E) comes out of the line as `one two`, collapsed rather than deleted, so neither can forge a record boundary or reorder what an operator reads. `LogSafeInvariantTest` derives its untrusted accessors from the response records and so does not reach a name read off a `JsonNode`; the name is wrapped where it is read, which is where that class's javadoc says the wrap belongs.

A `CHANGELOG.md` entry is added under `## [Unreleased]` → `### Fixed`.


devops-thiago:
@thrillhousebot resolved src/test/java/dev/thiagogonzaga/thrillhousebot/review/ai/FileSummariesDropDiagnosticsTest.java:131 — Sanitize test embeds a raw U+000A inside a JSON string, which a plain ObjectMapper rejects by default

Refuted: the mapper never sees that character raw. `ReviewResponseParser.parse` reads the document as `readDocuments(extractJson(raw))` (ReviewResponseParser.java:110), and `extractJson` ends on `return escapeControlCharsInStrings(trimmed)` (ReviewResponseParser.java:693), which escapes U+0000–U+001F inside string literals and leaves control characters outside them alone. That pass exists because models emit raw tabs and newlines inside string fields and strict parsing would fail the whole review; the test is deliberately taking the same route a model's own stray newline takes, and the name reaches the recovery as one<LF><RLO>two either way. The escape-sequence form would produce the identical in-memory name while skipping that step, which is why the raw form is the one worth testing. U+202E is a format character, not a control one, so it passes through untouched to LogSafe as intended. The full suite is green at 4042 tests locally and in CI. A comment at the line now says this, in 78455f9, so the choice is not read again as a test that only happens to pass.

Issue #735, opened 2026-08-16
Duplicate verdict ids drift between apply() and the coverage count, so an unverified finding is reported as fully screened


devops-thiago:
Found by deep-audit round 6 against the shipped v0.6.3. Introduced by #719 (rework) — the sibling shape of the #710 defect that fix closed.

#719's contract is that one decision normalizer feeds both `apply()` and the coverage count so the two cannot drift. That holds for a single verdict per id, but not for duplicates:

- `apply()` collapses duplicates first-wins and *then* reads the label (`FindingVerificationService.java:1186`, `:1195`)
- `candidatesCovered` filters by label *then* de-duplicates (`:757`, `:760`)

So a duplicate id whose **first** element is undecidable (missing/blank/unrecognized verdict) is kept unverified by `apply()` and counted as screened by the coverage pass: `(2,2)` → `Outcome.FULL` → `disclosed() == false` → no banner, no coverage clause, no check-run brief. That is exactly #710's harm — a published set containing an unscreened finding, reading as fully verified — one step narrower, and on a **complete, well-formed** response body rather than a cut one.

Executable proof: `probe1_duplicateIdDriftBetweenApplyAndCoverage` passes all four assertions; the control confirms the drift is one-directional (over-count, never under-count).

Fix direction: collapse duplicates identically in both readers (suggested patch in the audit report).

Related nit in the same file: `decisionOf` (`:770-774`) omits the `.strip()` that `strictRisk`/`strictConfidence` apply, contradicting its own javadoc at `:504-509`. Safe direction — a padded label reads as undecidable — but the inconsistency is worth closing with this.

Proof: audit report AUDIT6-B (findings F1, F3).

Issue #650, opened 2026-08-12
feat(review): resolve a finding's cited path:line against the real file before the verifier rules on it


devops-thiago:
> _Sanitized 2026-08-13: this ticket originally listed three directions. It is now committed to the **deterministic pre-check** (previously option 2), which its own incident history shows is sufficient; the others are recorded under Alternatives Considered. Part of the verification-correctness cluster in #669, alongside #475._

Split out of #646 (fixed by #649) and #636, both of which ran into the same limit from different directions.

### Problem Statement

The verifier is a single model call over fixed material with no ability to fetch anything. When a finding names a file and line the reviewer could not see in full, the verifier's only options are to accept it or reject it on the "not established by the provided material" ground — it cannot go and look. Its bias is to reject.

Where this has cost us:

- **#636** — a real security-invariant violation (`pr_purpose` spliced without `MarkdownSafe`) rejected because the quoted line was not in the batch's hunks. The line was real, nine lines from where the finding cited it. Went unpublished.
- **#589** — a correct Scala finding (`List.max` throws on empty) rejected as depending on "external library behavior not established by the diff".
- **#646** — artifact-reference findings rejected for depending on build-tool conventions; patched in the prompt by #649, but only if the build file happens to be in the material.

Each was band-aided by widening what the verifier may assume. Those prompt fixes rely on the producing evidence being inside the same window; when it is not, the verifier is back to guessing.

### Proposed Solution — deterministic cited-location resolution (no model call)

Before the verifier call, for each candidate finding that cites `path:line`:

1. Fetch the cited file's content at the head SHA (existing GitHub contents client; cache per review round — many findings cite the same files).
2. Locate the finding's quoted text in the file — **search for the quote first, cited line ± a small window as fallback** (#636's citation was 9 lines off; exact-line lookup alone would have missed it).
3. Attach the resolved snippet to that finding in the verifier material as established fact: "the file's actual content at/near the cited location is: …".
4. File unreachable or quote not found → attach nothing; the verifier behaves exactly as today. Strictly additive, fail-open.

A single resolved line would have flipped all three rejections above. Composes with #668 (per-finding verifier context): the resolved snippet rides with exactly the finding it belongs to.

### Acceptance criteria

- [ ] A finding citing a real line outside its batch's hunks (the #636 shape) is confirmed, not rejected, when the resolved content supports it
- [ ] Resolution is quote-search-first so near-miss line numbers still resolve
[...]

Issue #623, opened 2026-08-12
Verification silently does not run on ~1/3 of reviews


devops-thiago:
Finding verification fails often enough to matter, and when it does the review publishes unverified
findings with nothing saying so.

### Measured

Two hours of production logs on `test-sha-469539e-native`:

| outcome | count |
|---|---|
| `Finding verification: N kept, M downgraded, K rejected` (succeeded) | 10 |
| `Finding verification returned no response body — keeping the N unverified finding(s)` | 5 |
| `Finding verification response was cut mid-JSON` | 5 |

So roughly **one third of reviews publish findings that were never verified**. Both failure paths
fail *open*, which is the right direction — the findings still post rather than vanishing — and
neither is a crash. This issue is not about the fail-open behaviour.

### The problem

Nothing in the posted review distinguishes a verified finding set from an unverified one. The
maintainer sees the same comment either way, and so does anyone measuring the reviewer.

That matters concretely for the corpus measurements this project uses to decide what to fix. In
round 4, Scala (ThrillhouseBot-test #33) hit the empty-body path and published ten unverified
findings. It scored zero false positives — the same as every other language — but that number means
something different there: no second stage had screened it. The scorer only knew because the logs
were checked by hand and the fact was passed to it explicitly. Any future round without that manual
step will silently mix verified and unverified reviews into one precision figure.

Worth noting the round-4 result was interesting on its own terms — the raw, unverified findings were
already clean — but that is a conclusion the measurement should be able to state deliberately, not
one it stumbles into.

### Suggested direction

1. **Disclose it.** A review whose verification did not run should say so, in the same place other
   degradations are disclosed. The `SummaryDegradation` enum already models this shape (`NONE` /
   `RESPONSE_CUT` / `SKIPPED_AT_CEILING`) for the summary lane; verification could carry an
   equivalent.
2. **Investigate the rate.** Ten failures in two hours is not a rare edge. Both shapes point at the
   verifier's response rather than at the request: an empty body, and a body cut mid-JSON. The
   concise lane's caps and the reasoning-effort setting are the obvious things to look at first —
   #567 (concise lane inheriting reasoning effort) was fixed, so this is either a residue of that
   or something else.
3. **Consider a retry for the empty-body case specifically.** A single empty response is plausibly
   transient in a way a context-length rejection is not (see #622). Cheap to try once before giving
   up, unlike the 5x retry that issue objects to.

### Related

[...]

PR #874, opened 2026-09-16
fix(review): remember a rejected finding for the head it was rejected on

### LINKED ISSUE #711: The verifier reaches opposite verdicts on the same claim and unchanged code across runs
The finding verifier reaches opposite conclusions about the same claim on the same unchanged code,
within one dogfood round on one build. Two independent instances below.

## Instance 1 — a CRITICAL false positive rejected, then published

ThrillhouseBot-test #70, commit `2659f683`, unchanged between both reviews.

**Round 1** raised a SQL-injection finding on `ReservationRepository.cs` and the verifier rejected it:

> The URL path value is gated by the visible catalog allow-list (`rooms.FirstOrDefault(r => r.Code
> == code)`) before reaching the concatenated SQL, so injection would require a quote inside the
> trusted catalog, which is not shown and is outside attacker control.

That reasoning is correct. `FindByRoomAsync` has exactly one caller, `Program.cs:44`, three lines
below:

```csharp
app.MapGet("/rooms/{code}/availability", async (string code) =>
{
    var room = rooms.FirstOrDefault(r => r.Code == code);
    if (room is null) return Results.NotFound();          // gate
    var reservations = await repository.FindByRoomAsync(code, ...);
```

**Round 2**, ~40 minutes later on the same commit, published the same claim as 🔴 CRITICAL:

> FindByRoomAsync builds its SQL by string concatenation on this line, and roomCode flows in from the
> unauthenticated GET /rooms/{code}/availability route (Program.cs passes the route value straight
> into FindByRoomAsync).

"Straight into" is false — the catalog allowlist sits between them. The finding survived
verification this time.

The direction matters: the run that got it right suppressed it, and the run that got it wrong
published it at the highest severity, on a public route, where it is the kind of finding a maintainer
acts on immediately.

## Instance 2 — description gaps rejected on two PRs, kept on three

Same round, same build, five PRs each planting a PR-description-versus-code mismatch:

| PR | claim | verifier |
|---|---|---|
| go | retry/backoff described but not implemented | kept |
| kotlin | zero-cost rows described but never emitted | kept |
| zig | ignored pools described as staged, dropped instead | kept |
| rust | 409 described as carrying the earliest free slot, returns a fixed string | **rejected** |
| react | search described as matching submitter name, filters title only | **rejected** |

Both rejections cite the same ground — that the claim rests on a PR description absent from the
provided material:

> The claim that the 409 should carry an earliest-free-slot depends on a PR description not in the
> provided material, so the promised contract cannot be verified.

The code fact was equally checkable in all five: rust really does return a fixed string, react really
does filter on title only. The three that survived rested on the description exactly as much as the
two that did not.

## Why this is worth separating from ordinary model variance

A verifier that fails consistently shows up as a flat number every round and gets fixed. One that
fails half the time shows up as noise, and its errors land in whichever direction the sampling took —
here, a CRITICAL false positive on one side and correct findings deleted on the other.

It also undermines the measurement: five description-gap plants produced a 3/5 score that reflects
sampling rather than capability, and any future round will mix verified and unverified behaviour into
one figure without the logs to separate them.

Related: #475 (the verifier never sees the review context that produced the finding) is the
structural half of instance 2 — but the inconsistency shows the dependency is not absolute, so
supplying the context may fix the rejections without addressing the variance.

Found during round-7 dogfooding on v0.6.2 (`176371a`).

devops-thiago:
## What type of PR is this?

- [x] 🐛 Bug fix
- [ ] ✨ Feature
- [ ] 📝 Documentation
- [ ] 🔧 Refactor
- [ ] 🚀 Performance
- [ ] ✅ Test
- [ ] 🔒 Security
- [ ] 📦 Dependency update
- [ ] 🏗️ CI/CD

## Description

The second-pass audit reached opposite verdicts on one claim and one unchanged commit inside a
single dogfood round. On `ThrillhouseBot-test#70` at `2659f683` it rejected a SQL-injection
candidate with correct reasoning — the route value is gated by a catalog allow-list before it
reaches the concatenated query — and forty minutes later, on the same commit, published that same
claim as 🔴 CRITICAL on a public route.

### What #650 and #871 already fixed, and what is left

Both landed before this change, and both remove *reasons* the verifier was wrong. Neither touches
the fact that it answers the same question twice.

- **#650 (`CitedLocationResolver`)** resolves a finding's cited `path:line` against the real file at
  the head commit and attaches the result to the candidate. It closes the rejection ground "the
  evidence sits a few lines outside the nearest hunk, so nobody showed it to me".
- **#475, merged as #871 (`ContextEvidenceResolver`)** carries a context-grounded finding's own
  evidence — the patch-coverage measurement, the path-scoped rules — to the verifier. That is the
  structural half of the issue's instance 2: a description-versus-code claim was unverifiable by
  construction because the description was simply absent from the verifier's slots, and #871 gives
  the verifier the PR title and description (`FindingVerificationService.verify`'s `prContext`
  slot). The issue itself says supplying the context "may fix the rejections without addressing the
  variance", and that is exactly where things stand.

What is left is instance 1, and it is not a missing-material problem. The round that got the
SQL-injection claim right and the round that got it wrong were given the same code. Making the
inputs identical is not available as a fix either, for reasons the code settles rather than
guesses at:

- `FindingPipeline.refine`/`refineBatchOutcome` pass `promptInputs.previousFindings()` to the
  verifier, and that section grows with every posted round.
- The same call passes `PrContextBudget.bound(prContext, budgetPlanner.perCallInputBudget(), …)`,
  so what is truncated moves with the diff and the model's window.
- `ReviewContextLoader` reads the pull request conversation, which changes as people comment.

So two rounds on one head never see the same prompt, whatever the sampling settings are. This PR
makes the verdict stable by remembering it instead.

### The fix

`VerifierRejectionMemory` holds, per pull request and head commit, the candidates the audit threw
out. A later round on that head drops those claims in `FindingPipeline` **before** the verification
call, so they are neither re-litigated nor paid for.

- **Only rejections are remembered, never confirmations.** Re-auditing a confirmed finding costs a
  call; pinning one would repeat a false positive for the life of the head. The asymmetry is the
  safe direction — this store can only ever publish *fewer* findings than before.
- **A push clears the entry.** A rejection is an answer about code, and the code changed. The
  second test below is the control for exactly that.
- **A reworded re-raise is still the same claim.** Recall goes through
  `FollowUpAnalyzer.isSameFinding`, the rule the follow-up passes already use to recognize a
  re-raise, rather than a second recognizer that could disagree with them. That method was the only
  change to an existing class other than the pipeline wiring: `private` → package-private, with a
  javadoc saying why.
- **Caps and lifetime** match the neighbouring stores: 50 rejections per pull request (oldest
  forgotten first), 256 pull requests (least recently written evicted), in memory per replica, so a
  restart falls back to today's re-litigation — the same bound `CiHoldRegistry` and
  `SupersededFindingsCarryover` accept.

### The fail-open path

Yes, a fail-open path can publish a finding a successful verification would have dropped. The code
is explicit about it: in `FindingVerificationService.audit`, an empty body, any `IOException`/
`RuntimeException`, and the spend-ceiling refusal all `return screened` — the candidate posts
exactly as the reviewer raised it. That is the path behind the 16:00:29 UTC log line on our own
#871:

```
Finding verification returned no response body — keeping the 1 unverified finding(s)
```

The right disclosure already exists and is not changed here: #623's `VerificationCoverage` records
`(candidates, 0)` on every one of those paths and `VerdictBuilder` renders the clause — its own
javadoc records that production measured roughly one review in three publishing findings no second
stage had screened. What was missing is that the disclosure does not *drop* anything, so a claim the
audit had already rejected still posted. Applying the memory before the call fixes that: the drop
holds on a round that never gets a verdict at all. `aRejectedFindingIsNotRePublishedByALaterRoundOnTheSameHead`
stubs round two as the pass-through that every fail-open path produces, so it covers both the
"confirmed on the re-roll" and the "no verdict at all" shapes.

### Sampling — investigated, deliberately unchanged

Verified in code rather than assumed:

- `FindingVerifier` is bound to the `concise` named model, whose generation parameters are applied
  by `ChatModelCustomizers.ConciseChatModelCustomizer` through `BLOCKING_TUNING`.
- That customizer applies `temperature`, `topP`, `frequencyPenalty`, `presencePenalty` and `seed`
  from `ActiveModelSettings`, each only `ifPresent`. All five are unset in
  `application.properties` (the `thrillhousebot.ai.models.*.temperature` / `.seed` lines are
  commented-out examples), so the verifier call today runs at the provider's defaults with no seed.

So a deterministic setting *is* already available on the configured OpenAI-compatible provider
without any code change — `thrillhousebot.ai.models.<model>.temperature=0` and `.seed=<n>` reach the
verifier's lane. This PR does not flip that default, for two reasons. The effect cannot be proven
from a deterministic test — a test can show the builder received `temperature(0.0)`, which is wiring,
not behaviour — and the `concise` bean is shared with the summary, changelog and reply calls, where
pinning temperature is a prose-quality change nobody has measured. More to the point, greedy
decoding would not have fixed this issue: the two rounds did not send the same prompt, for the three
reasons listed above, so identical sampling would still have permitted different verdicts.

### Ordering, batch composition, truncation

Checked and found deterministic given the finding list, so nothing is changed there: candidate ids
are 1-based list positions in `renderCandidates`, `apply` reads them back by the same positions, and
`byCandidateId` collapses duplicate ids first-wins for both `apply` and `candidatesCovered` (#735).
Batch composition comes from `DiffBudgetPlanner`, which is a pure function of the file list and the
budget. The list order itself is model output and varies between rounds, which is one more reason
the stability has to come from memory rather than from input canonicalization.

## Related Issues

Fixes #711

## How Has This Been Tested?

- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

**Red first.** `FindingPipelineTest#aRejectedFindingIsNotRePublishedByALaterRoundOnTheSameHead`
runs two rounds against one `owner/repo#70` at `2659f683`; round one's verifier rejects the claim,
round two's returns its input unchanged (the shape both a `confirmed` verdict and every fail-open
path take). Against the unfixed code:

```
[ERROR] dev.thiagogonzaga.thrillhousebot.review.FindingPipelineTest.aRejectedFindingIsNotRePublishedByALaterRoundOnTheSameHead -- Time elapsed: 0.148 s <<< FAILURE!
org.opentest4j.AssertionFailedError: a claim rejected on this head must not be published by a later round on the same head ==> expected: <[]> but was: <[Concatenated SQL]>
	at dev.thiagogonzaga.thrillhousebot.review.FindingPipelineTest.aRejectedFindingIsNotRePublishedByALaterRoundOnTheSameHead(FindingPipelineTest.java:3154)
```

Its control, `aRejectionOnAnOlderHeadDoesNotSuppressTheFindingAfterAPush`, passes before and after:
it pins the boundary the memory must never cross.

`VerifierRejectionMemoryTest` covers the store itself — recall on the same head, recall of a
reworded re-raise, a finding the audit never rejected, the head moving, case-insensitive head
comparison, a push replacing rather than extending the entry, accumulation across rounds on one
head, a downgrade not read as a rejection, a finding citing no file, a session with no repository or
head, both caps. `ModelSuppliedTextInLogLinesTest#aCraftedPathAndTitleCannotForgeARecordFromTheRememberedRejectionDrop`
holds the new INFO line to the same log-forgery guarantee as its neighbours (#742/#755).

Gates, from the worktree root:

```
./mvnw -B clean compile spotbugs:check spotless:check
[INFO] BugInstance size is 0
[INFO] BUILD SUCCESS

./mvnw -B clean test
[INFO] Tests run: 4054, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

Patch coverage against `origin/main`, every added `src/main` line and branch:

```
GAPS: none
```

## Checklist

## Screenshots / Logs

The new drop is readable in the log, so a quiet round is never mistaken for a clean one:

```
Remembering 1 finding(s) the audit rejected for owner/repo#70 at 2659f683 (1 held); a later round on the same head will not put them to the verifier again
Dropping finding 'Concatenated SQL' (Repositories/ReservationRepository.cs:31): the second-pass audit rejected the same claim on head 2659f683 in an earlier round of this pull request
```

## Additional Notes

The memory is consulted and written in both review lanes — the single-call path in
`FindingPipeline.refine` and the per-batch path in `refineBatchOutcome` — so a large pull request
gets the same stability as a small one. Batches partition by file and recall requires the same file,
so one batch's rejection can never drop another batch's finding within a round.

What this does **not** do is make a wrong rejection right. If the audit rejects a true finding, this
change makes that rejection stick for the head instead of the finding reappearing on a later round.
That is the trade the issue asks for — "a finding verified and rejected on a head could be
remembered for that head rather than re-litigated from scratch" — and the reason #871 had to land
first: it removes the material gap that was producing the wrong rejections in the first place.

No new dependency, no configuration knob, no extra model call.


devops-thiago:
@thrillhousebot resolved src/main/java/dev/thiagogonzaga/thrillhousebot/review/VerifierRejectionMemory.java:140 — Javadoc says a fail-open round remembers nothing; the self-retraction screen's removals are remembered

Correct, and fixed in 6bf9856. The sentence contradicted the paragraph above it. `FindingVerificationService.audit` runs `demoteHedgedBlockingFindings(dropSelfRetractedFindings(response))` into `screened` before the model call, and all three fail-open paths — the empty body, the `IOException | RuntimeException` catch, and the spend-ceiling refusal — `return screened`. So on a round that lost its call, the published list is the screen's output, `removed(candidates, published)` is non-empty for anything the screen retracted, and the store does record it.

The javadoc now separates the two halves: the screen sits in front of the call and runs on every round, so its removals are recorded even when the call fails, which is a deterministic drop the next round would make again anyway; the verdict half contributes nothing on such a round, leaving a candidate the model never ruled on for a later round to rule on. Behaviour is unchanged — only the description was wrong.

